PR #5, opened 2020-08-07
introduce controller to watch openshift-config configmaps,secrets; upsert/delete configmaps/secrets in csi volume

gabemontero:
ok @adambkaplan @coreydaley here is the first pass at BUILD-56's acceptance criteria

so far, I've manually validated (which basic procedure noted in README)
- existence of `openshift-config` default set of configmaps and secrets in the csi volume of a user pod in another namespace
although not noted in the README specifically
- creation of new configmaps/secrets in `openshift-config` are picked up in the user pod
- deletion of those configmpas/secrets are picked up in the user pod
by additional `ls -lR` invocations while in the user pod

one might argue its "meets min" ... but let's start getting y'all's review

in the interim, possibly in a subsequent PR, 
- I'm going to start on some e2e/unit tests
- along with by a cross ref with https://github.com/kubernetes-sigs/secrets-store-csi-driver to possible augment the CSI interfaces with logic in that repo that "seems better"

sharing content from namespaces other than `openshift-config` will be (re)visited when we introduce the projected-resource group and types (i.e. `share`)


cmd/hostpathplugin/main.go
```diff
@@ -19,8 +19,11 @@ package main
 import (
 	"flag"
 	"fmt"
+	"github.com/openshift/projected-resource-csi-driver/pkg/controller"
```
coreydaley, line 22: Looks like the imports need sorted into sections here.

gabemontero: thanks update pushed

gabemontero:
> Is there a reason why this implementation has its own object cache? Why can't the `syncConfigMap` and `syncSecret` methods alter the filesystem themselves?

A few reasons:
- since we are not mutating the objects, we don't have to copy them when receiving them from the watches; and as we get pointer references, there is not a major memory difference per se; it is just an organizational / easy plug point for the csi driver itself
- we cannot mount in the same fs/volume as the csi volume that is shared; that would expose all the data to a given pod; we would have to create a separate hostpath volume to store this
- we don't get anything from a failure recovery perspective, since on restart the watches will repopulate the controller anyway
- we minimize the security exposure my minimizing how available the data is on the host  .... if a user pod is privileged, and if we cache the complete set of shared data on the host, the user pod can get access to data the cluster administrator may not want the particular pod to have via the projected resource share definition ... we only want the specific secrets/configmaps the projected resources shares should exist on the pod
- all the guidance from day one on this is to use tmp fs, tmp bind mounts, etc ... i.e. in-memory fs anyway ... why increase our use of the non-trivial means of setting all this up when we already have in-memory representations of the super set of configmaps/secrets we could possibly share 


It over complicates the solution with minimal, and a potential increase security exposure ... IMO

gabemontero:
@coreydaley @adambkaplan I've pushed my next round of iterative changes that includes
- pruning CSI plug points to the bare minimum needed
- picking desirable bits from https://github.com/kubernetes-sigs/secrets-store-csi-driver
- switching to in-memory tmpfs landing in the EP ... I'm leveraging k8s emptyDir tmpfs/in memory support re: https://kubernetes.io/docs/concepts/storage/volumes/#emptydir

pending iteration on this based on our review, my remaining items are
- post merge, review tmpfs specifics with storage SMEs ... there is a more complicated approach via the k8s Mounter which did not work for me ... I want to sanity check my use of memory based emtpydir with them vs. that approach, as well as see if there are additional options
- automated testing

Also, I think between @coreydaley and I, with perhaps some OLM help from @otaviof or @sbose78 , we need to figure out getting official images built for this vs. the ones currently pushed to my repo ... do we do that know, or just wait until the OLM work gets scheduled.  I'll bring this up in the next scrum.

gabemontero:
OK, pending resolution on building the image from openshift infra, I've pushed all the changes I plan for this sprints work and this PR

A couple of points:
1) the current volume and mount interactions reflect the feedback I got from the storage folks ... there is a detailed comment in the code and I'll update the jira story to relay those specifics
2) IMO the unit tests are meets min for sure 
3) I'm going to defer on the e2e test as I don't want to add all that manual deployment code in the e2e's that in theory should get replaces by official OLM deployment / image stuff  ... for now, the steps in the README for manual verification; once the deployment stuff is there, codifying the example pod spec and exec'ing into the running pod should suffice

Let's start diving into the review  @coreydaley @adambkaplan 

/assign @coreydaley 
/assign @adambkaplan 

thanks

pkg/hostpath/hostpath.go
```diff
@@ -202,6 +204,30 @@ func getVolumePath(volID string) string {
...
 }
 
+func commonUpsertRanger(path string, key, value interface{}) bool {
+	buf, err := json.MarshalIndent(value, "", "    ")
+	if err != nil {
+		glog.Errorf("error marshalling: %s", err.Error())
+		return true
+	}
+	filePath := filepath.Join(path, fmt.Sprintf("%s", key))
+	glog.V(0).Infof("GGM create/update file %s", filePath)
```
adambkaplan: side note - if we depend on k8s, we should switch to klog.

pkg/hostpath/constants.go
```diff
@@ -0,0 +1,28 @@
+/*
+Copyright 2017 The Kubernetes Authors.
```
adambkaplan, line 2: This code copied verbatim? If so, from where? Do we in fact need it?

gabemontero: yeah actually I created the file itself and then moved some of the constants / types shared between hostpath.go and nodeserver.go (and their test.go files) into it

I'll remove the k8s copyright

pkg/hostpath/hostpath.go
gabemontero: yeah for whatever reason https://github.com/kubernetes-csi/csi-driver-host-path uses `glog`

I'll handle the switch in a follow up PR right after this one merges

gabemontero:
update pushed @adambkaplan

PR #7, opened 2020-08-26
Clean up the Makefile and Dockerfiles

coreydaley:


gabemontero:
of course the driver yaml still points to my docker.io/gmontero @coreydaley ... my assumption was that was intentional until the new images actually exist on quay.io

coreydaley:
> of course the driver yaml still points to my docker.io/gmontero @coreydaley ... my assumption was that was intentional until the new images actually exist on quay.io

Correct.

PR #9, opened 2020-08-26
re-arranging a few things to fall in line with other repositories

coreydaley:


coreydaley:
@gabemontero I wonder if this is failing because I am moving the Dockerfile and ci doesn't have that update?  

` {"component":"entrypoint","file":"prow/entrypoint/run.go:169","func":"k8s.io/test-infra/prow/entrypoint.Options.ExecuteProcess","level":"error","msg":"Entrypoint received interrupt: terminated","severity":"error","time":"2020-08-26T16:31:50Z"} `

Chicken and the Egg?

PR #10, opened 2020-08-27
meets min repo rename adjustment

gabemontero:
might have a chicken/egg situation with openshift/release @adambkaplan @coreydaley and hence a green button situation

and meets min means it compiles, go mods, verifies, tests locally ... I say we consider changing driver names, namespace names, SA names, etc. separately


adambkaplan:
@bparees looks like prow is confused and I lost the green button. :frowning_face: 

cc @openshift/openshift-team-developer-productivity-test-platform

PR #12, opened 2020-08-27
fine tune copies; fix klog

gabemontero:
/assign @adambkaplan 

discovered the glog to klog port was not quite right, and it actually prevented the driver from coming up;
fixed that, along with verifying klog.V(0), klog.V(1), klog.V(2) up through klog.(5) work as expected now

also took another round of simplification of the host path / pod path mount flow

in particular:
- we just need to create the hostpath vol dir, we do not need to copy data there, since we have to copy the data to the pod tmpfs anyway because of our non-bind, 1 tmpfs per 1 pod, make selinux happy approach
- we do need to at least create the hostpath vol dir in its emptydir/medium: memory or else the mount fails with the non block access support stuff 

once we get consensus and merge this, I'll move onto

- move to the quay.io image ... mirroring should be there by then
- move on to projected resourece api

@coreydaley fyi

### DIFF
--- a/pkg/hostpath/nodeserver.go
+++ b/pkg/hostpath/nodeserver.go
@@ -150,11 +150,8 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 	klog.V(4).Infof("NodePublishVolume %v\nfstype %v\ndevice %v\nvolumeId %v\nattributes %v\nmountflags %v\n",
 		targetPath, fsType, deviceId, volumeId, attrib, mountFlags)
 
-	options := []string{} //"bind"}
-	//if readOnly {
-	//	options = append(options, "ro")
-	//}
-	path := ns.hp.getVolumePath(volumeId, podNamespace, podName, podUID, podSA)
+	options := []string{}
+	path := vol.VolPath
 
 	// NOTE: so our intent here is to have a separate tmpfs per pod; through experimentation
 	// and corroboration with OpenShift storage SMEs, a separate tmpfs per pod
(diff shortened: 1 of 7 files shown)


gabemontero:
uh oh @coreydaley 

can't assign @adambkaplan nor is a new PR of mine approved ??

coreydaley:
@gabemontero weird, it let me assign to you and myself, and i was able to approve it (shows you did also), it just didn't apply the label ...

coreydaley:
> uh oh @coreydaley
> 
> can't assign @adambkaplan nor is a new PR of mine approved ??

Looks like it is working again, maybe a ci hiccup?

adambkaplan:
Note - I'm OK moving to klog v2 in a separate PR.

pkg/hostpath/nodeserver.go
```diff
@@ -166,8 +163,8 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 	// The various bits that work in concert to achieve this
 	// - the use of emptyDir with a medium of Memory in this drivers Deployment is all that is needed to get tmpfs
 	// - do not use the "bind" option, that reuses existing dirs/filesystems vs. creating new tmpfs
-	// - without bind, we have to specify an fstype of tmpfs, or we get errors on the Mount about the fs not being
-	//   block access
+	// - without bind, we have to specify an fstype of tmpfs, and path for the mount source, or we get errors on the
```
adambkaplan, line 166: nit - extra comma:

```suggestion
	// - without bind, we have to specify an fstype of tmpfs and path for the mount source, or we get errors on the
```

gabemontero:
updates pushed @adambkaplan including klog v2

PR #13, opened 2020-08-28
Update to use the new naming scheme and quay.io images

coreydaley:


coreydaley:
We should probably let this one site until @gabemontero is back and let him check/test it.

deploy/csi-hostpath-plugin.yaml
```diff
@@ -60,13 +60,13 @@ spec:
               name: csi-data-dir
 
         - name: hostpath
-          image: docker.io/gmontero/origin-projected-resource-csi-driver:latest
+          image: docker.io/gmontero/origin-csi-driver-projected-resource:latest
```
adambkaplan, line 63: Don't we want to reference the quay image here?

Makefile
```diff
@@ -28,8 +28,8 @@ build: ## Build the executable. Example: make build
 
 build-image: ## Build the images and push them to the remote registry. Example: make build-images
 	rm -rf _output
-	docker build -f Dockerfile -t $(REGISTRY)/$(REPOSITORY)/origin-projected-resource-csi-driver:$(TAG) .
-	docker push $(REGISTRY)/$(REPOSITORY)/origin-projected-resource-csi-driver:$(TAG)
+	docker build -f Dockerfile -t $(REGISTRY)/$(REPOSITORY)/origin-csi-driver-projected-resource:$(TAG) .
+	docker push $(REGISTRY)/$(REPOSITORY)/origin-csi-driver-projected-resource:$(TAG)
```
adambkaplan, line 32: We should let the container runtime be configurable, so folks can use `podman` without the docker alias.

Or crazy people like me, who need to do `flatpak-spawn --host podman`:

Up top:
```
CONTAINER_RUNTIME ?= docker
```
Then here:

```
$(CONTAINER_RUNTIME) build ...
$(CONTAINER_RUNTIME) push ...
```

deploy/csi-hostpath-plugin.yaml
coreydaley: Yep, i missed that one, thanks.

Makefile
coreydaley: updated in current push

coreydaley:
@adambkaplan updated per your comments, ptal

adambkaplan:
/hold

I think we just need to verify the quay image is public.

coreydaley:
/hold cancel
The quay image is public.

gabemontero:
I'll checkout this PR after #12 merges and the images is updated ... need #12 to merge to get non-default logging

gabemontero:
OK I see the image is now updated on quay with #12 and I've checked out this PR's branch @coreydaley 

```
gmontero ~/go/src/github.com/openshift/csi-driver-projected-resource  (local-upstream-13-copy)$ git log -1
commit 8806faf1dfd005e8e2b9fb6b1ef2acf36a767a10 (HEAD -> local-upstream-13-copy)
Author: Corey Daley <cdaley@redhat.com>
Date:   Fri Aug 28 09:59:00 2020 -0400

    Update to use the new naming scheme and quay.io images

```

My cluster for today has been destroyed though.  I'll give this a go first thing in the AM when I boot up a new cluster.

gabemontero:
/lgtm

local testing confirmed

PR #14, opened 2020-09-01
Implement cobra for command line

coreydaley:


coreydaley:
@gabemontero ptal

main.go
```diff
@@ -0,0 +1,9 @@
+package main
```
gabemontero, line 1: Yeah i'm not fan of main.go in the top of the repo

move it to the cmd folder and employ filenames similar to https://github.com/openshift/builder/tree/master/cmd

coreydaley: updated in current push, ptal

coreydaley:
/hold
Should probably wait for https://github.com/openshift/csi-driver-projected-resource/pull/15 to merge and then I can rebase this one if needed.

Makefile
```diff
@@ -31,7 +31,7 @@ verify: ## Run verifications. Example: make verify
 .PHONY: verify
 
 build: ## Build the executable. Example: make build
-	go build -a -race -ldflags $(LDFLAGS) -o _output/hostpathplugin ./cmd/hostpathplugin
+	go build -a -mod=vendor -race -ldflags $(LDFLAGS) -o csi-driver-projected-resource ./cmd
```
gabemontero, line 34: if you put it here, then we have to deal with our .gitignore and the new binary

whereas the _output dir is a convention we already have in place for openshift

gabemontero (changes requested):
one minor item on binary location

Makefile
coreydaley: update pushed.

PR #16, opened 2020-09-10
BUILD-57: finallize the projected resource API

gabemontero:
/assign @adambkaplan 
/assign @coreydaley 
@otaviof FYI

Between the commit text and README update, hopefully a description of what is getting achieved with this PR.

I believe this gets us to meets min for this sprint's story.  But let's discuss.

Per prior discussion with @adambkaplan I did not take on an even more precise approximation of https://kubernetes.io/docs/concepts/storage/volumes/#projected

But if desired we can dive into that as well.

PTAL


pkg/controller/controller.go
```diff
@@ -50,42 +54,103 @@ func NewController() (*Controller, error) {
...
+	shareClient, err := shareclientv1alpha1.NewForConfig(kubeRestConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	// NOTE, not specifying a namespace defaults to metav1.NamespaceAll in
+	// informers.NewSharedInformerFactoryWithOptions, but we restrict OpenShift
+	// "system" namespaces with chatty configmaps like the leaderelection related ones
+	// that are updated every few seconds
+	tweakListOptions := internalinterfaces.TweakListOptionsFunc(func(options *metav1.ListOptions) {
```
gabemontero, line 66: So in my testing, this function gets called more than once.  But on prolonged intervals, and intervals which I could not line up with the relist interval.

In theory, we could have flipped the field selector string to list namespaces we are interested in based on the shares we discover, but given this was called so infrequently, I deemed that approach untenable.

pkg/hostpath/hostpath.go
```diff
@@ -163,52 +177,195 @@ func commonUpsertRanger(podPath string, key, value interface{}) bool {
...
+	oldTargetPath := ""
+	volID := ""
+	change := false
+	lostPermissions := false
+	gainedPermissions := false
+	hpv := &hostPathVolume{}
+	for _, hpv = range hostPathVolumes {
+		if hpv.SharedDataId == shareId {
+			klog.V(4).Infof("share update ranger id %s found volume %s", shareId, hpv.VolID)
+			a, err := client.ExecuteSAR(shareId, hpv.PodNamespace, hpv.PodName, hpv.PodSA)
```
gabemontero, line 235: So this is effectively how we check if a pod still has permissions for its share.  Since it is based on the share update, it works off of the share relist interval, so it is a "lazy check".

I did this vs. watching all the ClusterRoleBindings and RoleBindings and then deciphering which were related to the shares currently being consumed by Pods.

Making the relist interval for the share watch configurable has occurred to me as a way to accelerate things.  But I have not added it as of yet.

Seems like it should be safe, given we should not expect an abundance of this per se, but I'm interested in opinions.

README.md
```diff
@@ -22,15 +22,61 @@ Enhancement proposal.
 
 **NOT FULLY IMPLEMENTED**
 
-The latest commit of the master branch solely brings in the subset of the hostpath driver needed to validate 
-the basic in-line ephemeral volume scenario articulated below. 
-
+The latest commit of the master branch solely introduces both the `Share` CRD and the `projectedresoure.storage.openshift.io`
```
gabemontero, line 25: Aside from fallout from the review of this PR, as well as the possible evolution at some point of the API closer toward https://kubernetes.io/docs/concepts/storage/volumes/#projected the other major items currently in my mind for follow on work:

1) persist the csi driver volume map to disk so that it can be re-populated after the driver has been recycled
2) the OLM install stuff
3) in conjunction, e2e tests

pkg/hostpath/hostpath.go
gabemontero: The share relist interval is now configurable as a container arg

README.md
gabemontero: I've now pushed changes for "persist the csi driver volume map to disk so that it can be re-populated after the driver has been recycled"

Given what we came up in planning today, items 2) and 3) will be handled in subsequent stories / PRs @adambkaplan @coreydaley @otaviof 

Also per planning we can discuss the current approach toward permissions here, but most likely will defer subsequent changes to other stories/PRs

PTAL at this one when you get the chance

pkg/hostpath/nodeserver.go
```diff
@@ -65,22 +72,51 @@ func getPodDetails(volumeContext map[string]string) (string, string, string, str
...
 	}
+	return nil, err
+}
+
+// validateVolumeContext return values:
+// - podNamespace
+// - podNames
+// - podUID
+// - podSA
+func (ns *nodeServer) validateVolumeContext(req *csi.NodePublishVolumeRequest) (string, string, string, string, error) {
```
adambkaplan, line 119: Why do we need to return the pod data when this information exists in the `NodePublishVolumeRequest`?

Returning four `strings` plus an `error` feels like a golang code smell.

adambkaplan (changes requested):
Mostly nits

gabemontero:
re: `internalinterfaces` @adambkaplan ..... it is only referenced because it is explicitly cited it in a public  API in a "public" package:

from k8s.io/client-go/informers

```
// WithTweakListOptions sets a custom filter on all listers of the configured SharedInformerFactory.
func WithTweakListOptions(tweakListOptions internalinterfaces.TweakListOptionsFunc) SharedInformerOption {
	return func(factory *sharedInformerFactory) *sharedInformerFactory {
		factory.tweakListOptions = tweakListOptions
		return factory
	}
}
```

An oversight on k8s's part IMO.  And from my investigation this is the only facility that will allow us to reasonably achieve what we need.

pkg/controller/controller.go
```diff
@@ -50,42 +54,104 @@ func NewController() (*Controller, error) {
...
+			"openshift-config-operator",
+			"openshift-etcd-operator",
+			"openshift-apiserver-operator",
+			"openshift-cluster-csi-drivers",
+			"openshift-cluster-storage-operator",
+			"openshift-cluster-version",
+			"openshift-image-registry",
+			"openshift-machine-config-operator",
+			"openshift-sdn",
+			"openshift-service-ca-operator",
```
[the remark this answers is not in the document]
gabemontero: part of next push

pkg/hostpath/nodeserver.go
gabemontero: what, multiple returns is one of the joys of golang :-)

I'll make an attempt at restructuring to this to facilitate the various concerns, including what you've noted, and see if it sticks at least for me.

gabemontero: OK it accesses the vol ctx an time or two more than before, but the organization nodeserver vs. hostpath is much better, and among other things this methods returns just an error now @adambkaplan 

thanks

gabemontero:
Updates to address comments pushed @adambkaplan PTAL

adambkaplan:
> An oversight on k8s's part IMO. And from my investigation this is the only facility that will allow us to reasonably achieve what we need.

Can't argue with that - is there an issue upstream?

gabemontero:
> > An oversight on k8s's part IMO. And from my investigation this is the only facility that will allow us to reasonably achieve what we need.
> 
> Can't argue with that - is there an issue upstream?

There is a rework RFE out there that would address this, though it is fighting rotten lifecycle

https://github.com/kubernetes/client-go/issues/801

pkg/hostpath/nodeserver.go
```diff
@@ -189,6 +244,7 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 			err.Error()))
 	}
 
+	storeVolMapToDisk()
```
adambkaplan, line 247: Check for error and raise here.

adambkaplan (changes requested):

pkg/hostpath/nodeserver.go
gabemontero: update for this requested change pushed @adambkaplan

pkg/hostpath/nodeserver.go
```diff
@@ -244,7 +244,11 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 			err.Error()))
 	}
 
-	storeVolMapToDisk()
```
coreydaley: just a nit
```
	if err = storeVolMapToDisk(); err != nil {
		klog.Errorf("failed to persist driver volume metadata to disk: %s", err.Error())
		return nil, status.Error(codes.Internal, err.Error())
	}
```

gabemontero: update for this pushed @coreydaley

adambkaplan:
@gabemontero squash commits then I think we're good.

gabemontero:
> @gabemontero squash commits then I think we're good.

commits squashed @adambkaplan thanks

adambkaplan (approved):
/lgtm

PR #17, opened 2020-10-28
BUILD-169: change filesystem storage approach to mirror k8s/k8s secret/configmap volumes

gabemontero:
/assign @coreydaley-redhat 

manual testing looks good  ... some of which is reflected in the README update,
though I went through the set of scenarios at https://github.com/openshift/csi-driver-projected-resource#current-status-with-respect-to-the-enhancement-propsal

### DIFF
--- a/pkg/hostpath/hostpath.go
+++ b/pkg/hostpath/hostpath.go
@@ -18,20 +18,22 @@ package hostpath
 
 import (
 	"encoding/gob"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
 	"sync"
 
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/klog/v2"
+
 	sharev1alpha1 "github.com/openshift/csi-driver-projected-resource/pkg/api/projectedresource/v1alpha1"
 	objcache "github.com/openshift/csi-driver-projected-resource/pkg/cache"
 	"github.com/openshift/csi-driver-projected-resource/pkg/client"
-
-	"k8s.io/klog/v2"
 )
 
 const (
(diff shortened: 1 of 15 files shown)


pkg/hostpath/hostpath.go
```diff
@@ -18,10 +18,11 @@ package hostpath
 
 import (
 	"encoding/gob"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
+	corev1 "k8s.io/api/core/v1"
```
coreydaley-redhat, line 25: Just a nit, imports are not organized.

pkg/hostpath/hostpath.go
```diff
@@ -166,28 +167,38 @@ func (hp *hostPath) getVolumePath(volID string, volCtx map[string]string) string
...
 		return true
 	}
-	buf, err := json.MarshalIndent(value, "", "    ")
-	if err != nil {
-		klog.Errorf("error marshalling: %s", err.Error())
+	payload, _ := value.(Payload)
+	podFileDir := filepath.Join(podPath, fmt.Sprintf("%s", key))
+	if err := os.MkdirAll(podFileDir, os.ModePerm); err != nil {
+		klog.Errorf("unable to make dir %s: %s", podFileDir, err.Error())
 		return true
```
coreydaley-redhat, line 178: Should this return false since it's a failure?

pkg/hostpath/hostpath.go
```diff
@@ -166,28 +167,38 @@ func (hp *hostPath) getVolumePath(volID string, volCtx map[string]string) string
...
-	podFilePath := filepath.Join(podPath, fmt.Sprintf("%s", key))
-	klog.V(4).Infof("create/update file %s", podFilePath)
-	createFile(podFilePath, buf)
+	if payload.ByteData != nil {
+		for dataKey, dataValue := range payload.ByteData {
+			podFilePath := filepath.Join(podFileDir, dataKey)
+			klog.V(4).Infof("create/update file %s", podFilePath)
+			if err := ioutil.WriteFile(podFilePath, dataValue, 0644); err != nil {
+				klog.Errorf("unable to write file %s: %s", podFilePath, err.Error())
+				return true
```
coreydaley-redhat, line 186: Should return false here?

pkg/hostpath/hostpath.go
```diff
@@ -166,28 +167,38 @@ func (hp *hostPath) getVolumePath(volID string, volCtx map[string]string) string
...
+		}
+	}
+	if payload.StringData != nil {
+		for dataKey, dataValue := range payload.StringData {
+			podFilePath := filepath.Join(podFileDir, dataKey)
+			klog.V(4).Infof("create/update file %s", podFilePath)
+			content := []byte(dataValue)
+			if err := ioutil.WriteFile(podFilePath, content, 0644); err != nil {
+				klog.Errorf("unable to write file %s: %s", podFilePath, err.Error())
+				return true
```
coreydaley-redhat, line 198: Should return false here?

pkg/hostpath/hostpath.go
```diff
@@ -166,28 +167,38 @@ func (hp *hostPath) getVolumePath(volID string, volCtx map[string]string) string
...
 	}
-	podFilePath := filepath.Join(podPath, fmt.Sprintf("%s", key))
-	klog.V(4).Infof("create/update file %s", podFilePath)
-	createFile(podFilePath, buf)
+	if payload.ByteData != nil {
+		for dataKey, dataValue := range payload.ByteData {
+			podFilePath := filepath.Join(podFileDir, dataKey)
+			klog.V(4).Infof("create/update file %s", podFilePath)
+			if err := ioutil.WriteFile(podFilePath, dataValue, 0644); err != nil {
+				klog.Errorf("unable to write file %s: %s", podFilePath, err.Error())
```
coreydaley-redhat, line 185: err.Error() here probably includes the file path that it could not write to, what about just returning an error from this (along with the similar ones below) and handling that error where the function is called?  Seems like that might be cleaner and less repetitive code of calling klog to print the error and then returning false?

pkg/hostpath/hostpath.go
```diff
@@ -339,7 +350,12 @@ func mapBackingResourceToPod(hpv *hostPathVolume) error {
 			return err
 		}
 		upsertRangerCM := func(key, value interface{}) bool {
-			return commonUpsertRanger(podConfigMapsPath, hpv.SharedDataKey, key, value)
+			cm, _ := value.(*corev1.ConfigMap)
+			payload := Payload{
+				StringData: cm.Data,
+				ByteData:   cm.BinaryData,
+			}
+			return commonUpsertRanger(podConfigMapsPath, hpv.SharedDataKey, key, payload)
```
coreydaley-redhat, line 358: Here you could handle any errors returned by the commonUpsertRanger function and return true/false and log the error.

coreydaley-redhat (changes requested):

pkg/hostpath/hostpath.go
gabemontero: Yeah this stems from how the golang ranger contract works.... there is not error propagation.

To date I've been of the opinion that best effort is sufficient, and that if we are getting golang filesystem errors like this, most likely retry is not going to help, as most likely we are hitting "the host is messed up" errors vs. intermittent RPC over the network type things.  So kind of "special casey".

And I like utilizing golang based API here.

We'll ultimately get a retry when the controller relists the configmaps/secrets anyway.... though at a much more gradual interval.

That said, I'll think about it some more, and see if I'm convinced we should pivot at all.  Minimally, I should add comments to capture all this nuance.

pkg/hostpath/hostpath.go
gabemontero: Yeah this will capture at least errors on initial creations.  If we get errors on subsequent updates after the Pod's volume has been provisioned and the configmap/secrets have been updated, this path won't help.

For that, I'm going to add k8s event recording to at least augment the the error logging in the pod.

gabemontero:
updates pushed @coreydaley-redhat PTAL thanks

pkg/hostpath/hostpath.go
```diff
@@ -166,29 +168,51 @@ func (hp *hostPath) getVolumePath(volID string, volCtx map[string]string) string
...
+	// Then, the controller relist will result in going through the secrets/configmaps we share, so
+	// again, on the off chance the filesystem error is intermittent, or if an administrator has taken corrective
+	// action, writing the content will be retried.  And note, the relist interval is configurable (default 10 minutes)
+	// if users want more rapid retry...but by default, no tight loop more CPU intensive retry
+	// Lastly, with the understanding that an error log in the pod stdout may be missed, we will also generate a k8s
+	// event to facilitate exposure
+	// TODO: prometheus metrics/alerts may be desired here, though some due diligence on what k8s level metrics/alerts
+	// around host filesystem issues might already exist would be warranted with such an exploration/effort
+	if err := os.MkdirAll(podFileDir, os.ModePerm); err != nil {
+		msg := fmt.Sprintf("unable to make dir %s: %s", podFileDir, err.Error())
```
coreydaley-redhat, line 190: Did you mean to use your new ProcessFileSystemError function here?

gabemontero: fixed to use new function @coreydaley-redhat thanks

PR #20, opened 2021-01-20
fix hpv data race; employ standard controller event copy

gabemontero:


gabemontero:
/skip

Yep ... now realize that our e2e's with the deploy script are using the current master branch image pushed to quay and not the PR branch image stored in the CI pipeline imagestream at 

image-registry.openshift-image-registry.svc:5000/<PR test name space>/pipeline:csi-driver-projected-resource

To get it to work, we would need to change the images used in the DaemonSet defined in deploy/csi-hostpath-plugin.yaml to point that image assoicated with the imagestream tag

Either I try to do some short term solution with this current script approach, or we wait for @coreydaley 's OLM changes, which presumably would provide the equivalent use of the PR's produced images for testing like we get with our other https://github.com/openshift repos

gabemontero:
> /skip
> 
> Yep ... now realize that our e2e's with the deploy script are using the current master branch image pushed to quay and not the PR branch image stored in the CI pipeline imagestream at
> 
> image-registry.openshift-image-registry.svc:5000//pipeline:csi-driver-projected-resource
> 
> To get it to work, we would need to change the images used in the DaemonSet defined in deploy/csi-hostpath-plugin.yaml to point that image assoicated with the imagestream tag
> 
> Either I try to do some short term solution with this current script approach, or we wait for @coreydaley 's OLM changes, which presumably would provide the equivalent use of the PR's produced images for testing like we get with our other https://github.com/openshift repos

OK I got it to work to use the image produced by a given PR's changes.  We'll see how @coreydaley 's Jira shakes out, but I do recall after (re)discovering `IMAGE_FORMAT` that it was originally created for non-payload repo's (like our soon to be OLM deployed stuff here)

coreydaley:
@gabemontero I have a question, partially about this pull request, but sort of about the way you are passing data around as a whole.

It seems to me like you are using the sync.Map as a queue. You put things into the sync.Map, process them, and then remove them. If that is the case, wouldn't it make more sense to use channels? You could then have multiple workers (go routines) process the queues as you add items to them in the order that they were added, even going so far as to have multiple workers per queue. You also would not be storing the extra data (the key) that is needed for a map. I'm not sure how many configmaps are being processed at one time, so maybe that doesn't matter as much.

If I remember correctly, the entries in a map are not guaranteed to be in a specific order, which may not matter much unless things are being added as the map is being processed?

It is also possible that I am misinterpreting the usage and you are, in fact, writing back to the objects in the sync.Map, in which case what you are currently doing would make sense.

gabemontero:
> @gabemontero I have a question, partially about this pull request, but sort of about the way you are passing data around as a whole.
> 
> It seems to me like you are using the sync.Map as a queue. You put things into the sync.Map, process them, and then remove them. If that is the case, wouldn't it make more sense to use channels? You could then have multiple workers (go routines) process the queues as you add items to them in the order that they were added, even going so far as to have multiple workers per queue. You also would not be storing the extra data (the key) that is needed for a map. I'm not sure how many configmaps are being processed at one time, so maybe that doesn't matter as much.
> 
> If I remember correctly, the entries in a map are not guaranteed to be in a specific order, which may not matter much unless things are being added as the map is being processed?
> 
> It is also possible that I am misinterpreting the usage and you are, in fact, writing back to the objects in the sync.Map, in which case what you are currently doing would make sense.

So @coreydaley reached out to me on slack as well and we ended up talking there a bit.  I'm going to attempt to capture the net, then provide needed explanation on the code.  So this ended up being tl'dr'ish but hopefully that is not a barrier to it being useful.

1) He is going to take another pass at the code in general vs. these code changes in particular, and then we'll reset
2) minimally, additional comments in the code on the intent of things seems warranted
3) At a high level the answer to his question `you are using the sync.Map as a queue. You put things into the sync.Map, process them, and then remove them.` is "no".  And entries existence in a Map lives beyond the processing of an event.  Entries only get removed as the associated shares, secrets, and configmaps are deleted in the cluster.
If anything, the maps are more analogous with database tables, with "queries" represented by how the `Range` function is used (where we create "maps who are subsets of other maps)
4) I'm going to make a first pass at writing some higher level description of things here ... if I'm successful in conveying things, this will translate into comments in the code

Cardinality:

- 1 share can be referenced by many pods, 1 share references either a secret or configmap
- so that means 1 secret or configmap can be stored in many pods

Event type (with some timing nuances) / needed action:

- share created before referenced config map / secret: any pod created citing the shares should fail on volume mount, but kubelet will retry some, so if the configmap / secret arrive after the share, but within that window, things should work, with the secret/configmap stored in the pod's csi volume
- share created after referenced config map / secret: any pod created citing the share should come up with the secret/configmap contents in the csi volume
- share updated (switch to different secret/configmap, switch from secret to configmap): the old referenced data needs to be removed, the new referenced data needs to be added
- share deleted: assuming any long running pods reference the share, the associated csi volume needs to be removed from those pod
- permissions for pod added before share/pod created: SAR check should pass, data stored
- permission for pod added after share/pod create: if within the kubelet's retry window, SAR check eventually passes, data stored; otherwise, pod fails mount
- secret / configmap updated: each long running pod with a share ref'ing those need to have their csi volume contents updated
- secret / configmap deleted: each long running pod with a share ref'ing those needs to have their csi volume contents removed
- pod with share shutdown: clean up storage on host as needed, update or Maps containing the share/configmap/secret to pod refs

Data structures to facilitate event processing:

So a lot traversal paths other either the entire set or a subset of shares / configmaps / secrets.  The code that facilitates those
leverages golang synchronous Maps in large part because the Range method on those Maps is based on providing a function to be called as each entry in the Map is visited, where
1) part of the input to the function is the key value pair of the map being ranged over
2) or input to the function could arbitrary data needed by the function can be stored as functional variables, where that data is available when the function is created
3) so that we could prep needed update / delete functions when items are initially registered
4) and then subsequent events simply range over maps

So, what maps do we create
a) 1 map for all shares (cluster scoped) indexed by the share ID (ref to this in the pod)
(master in memory list code can use for easy access, cross referencing, and ranging on the "submaps" below)
b) 1 map for all configmaps (namespace scoped) indexed by a key with namespace and name (ref to this in the share)
(master in memory list code can use for easy access, cross referencing, and ranging on the "submaps" below)
c) 1 map for all secrets same as ^^
(master in memory list code can use for easy access, cross referencing, and ranging on the "submaps" below)
d) secret/configmap maps with a key of the volume ID and value of the secret/share
(as pods go away and volume are removed we can remove the temp file systems associated with secrets or configmaps when no more pods reference the given secret/configmap)
e) 1 map for secrets who have shares referencing them
(so we by default can ignore secrets who are not referenced by any shares)
f) 1 map for configmaps who have shares referencing them
(so we by default can ignore configmaps who are not referenced by any shares)
f) 1 map for a share who references a secret we have not received an event for yet
 (these were added in this PR as part of fixing the bug where on daemonset restart share events came before configmap/secret events)
g) 1 map for a share who references a configmap we have not received an event for yet 
(these were added in this PR as part of fixing the bug where on daemonset restart share events came before configmap/secret events)
h) maps to hold functions to be called by Map.Range when a give event comes into the controller.

The controller event processing boils down to traversing the required subset of objects associated with the event object to execute the action (captured by a function previously supplied to be used in the Map Range function)

gabemontero:
ok @coreydaley I've pushed the comments previously mentioned ... let me know what you think

gabemontero:
ci flake with slow

/test e2e-aws-operator-slow

PR #22, opened 2021-02-10
fix delete of configmap/secret with keys

gabemontero:
the unit tests were not creating actual files resulting from k/v's in the configmap/secret data maps

hence the need for an `os.RemoveAll` vs. the existing `os.Remove` was missed

also did some cleanup around better unit test concurrency and string literals

### DIFF
--- a/pkg/hostpath/hostpath.go
+++ b/pkg/hostpath/hostpath.go
@@ -216,7 +216,7 @@ func commonDeleteRanger(podPath, filter string, key interface{}) bool {
 		return true
 	}
 	podFilePath := filepath.Join(podPath, fmt.Sprintf("%s", key))
-	os.Remove(podFilePath)
+	os.RemoveAll(podFilePath)
 	return true
 }
 
--- a/pkg/hostpath/hostpath_test.go
+++ b/pkg/hostpath/hostpath_test.go
@@ -35,16 +35,23 @@ import (
 	"github.com/openshift/csi-driver-projected-resource/pkg/client"
 )
 
-func testHostPathDriver() (*hostPath, string, string, error) {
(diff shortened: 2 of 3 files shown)


gabemontero:
Another data race in the disruptive test .. this time it was 2 separate configmap/secret events trying to init the client:

```
        ==================
        WARNING: DATA RACE
        Write at 0x0000030fa510 by goroutine 81:
          github.com/openshift/csi-driver-projected-resource/pkg/client.initClient()
              /go/src/github.com/openshift/csi-driver-projected-resource/pkg/client/client.go:88 +0x279
          github.com/openshift/csi-driver-projected-resource/pkg/client.ExecuteSAR()
              /go/src/github.com/openshift/csi-driver-projected-resource/pkg/client/client.go:93 +0x47
```

will see how pervasive it is with a few more runs.  If pervasive, I'll have to fix here for merge, otherwise will fix separately.

gabemontero:
ok it passed this time ... let's try to review / merge this change as is @otaviof and I'll deal with the data race in a separate PR

otaviof (approved):
@gabemontero The PR looks good, aside of the `os.RemoveAll` you've made good improvements in the testing code!

otaviof:
A nit comment about `hostpath.primeSecretVolume` signature, currently we have:

https://github.com/openshift/csi-driver-projected-resource/blob/2ba5bb4acf3aef39584906431ef5d2a315d73da8/pkg/hostpath/hostpath_test.go#L442

However, usually `t` (testing) comes as a first argument, and by having `t`, it is always able to extract the test name (`t.Name()`). Therefore, we can simplify the method signature as the following snippet:

```go
func primeSecretVolume(t *testing.T, hp *hostPath, targetPath, share *sharev1alpha1.Share) *corev1.Secret
```

PR #24, opened 2021-02-12
fix removal of files when associated keys are removed

gabemontero:
with adding k/v's to the configmap/secret unit tests, also discovered handling of changes in which keys were removed in a configmap/secret was not being properly handled

the first commit addresses that, and adds a unit test

the second commit addresses @otaviof 's https://github.com/openshift/csi-driver-projected-resource/pull/22#issuecomment-778026663

/assign @otaviof


otaviof:
The changes here are looking good, and thanks for taking the previous comment in consideration!

/lgtm

gabemontero:
> The changes here are looking good, and thanks for taking the previous comment in consideration!
> 
> /lgtm

likewise thanks @otaviof

PR #27, opened 2021-02-19
remove type/key subpath on mount

gabemontero:
/assign @adambkaplan 

per prior discussions, removing the kind/namespace:name subdirs 

in the end, I've confirmed they provided no value, even if one share is mounted as a subdir of another share, 
so I nixed the notion of enabling both behaviors

with this, in our entitlements demo, we can mount directly into /etc/pki/entitlement, as with the vanilla ubi8 image, there is not content by default in that dir.

### DIFF
--- a/pkg/hostpath/hostpath.go
+++ b/pkg/hostpath/hostpath.go
@@ -173,7 +173,6 @@ func commonUpsertRanger(obj runtime.Object, podPath, filter string, key, value i
 	}
 	payload, _ := value.(Payload)
 	klog.V(4).Infof("common upsert ranger key %s", key)
-	podFileDir := filepath.Join(podPath, fmt.Sprintf("%s", key))
 	// So, what to do with error handling.  Errors with filesystem operations
 	// will almost always not be intermittent, but most likely the result of the
 	// host filesystem either being full or compromised in some long running fashion, so tight-loop retry, like we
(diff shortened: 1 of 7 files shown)


pkg/hostpath/hostpath.go
```diff
@@ -221,13 +220,28 @@ func commonUpsertRanger(obj runtime.Object, podPath, filter string, key, value i
...
+func commonOSRemove(dir string) error {
+	// we cannot do a os.RemoveAll on the mount point, so we remove all on each file system entity
+	// off of the potential mount point
+	return filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
+		if info == nil {
+			return nil
+		}
+		// since we do not support mounting on existing content, a dir can only mean a share
+		// has been mounted as a separate dir in our share, so skip
+		if info.IsDir() {
```
gabemontero, line 232: FYI, the e2e `TestTwoSharesSeparateButInheritedMountPathsRemoveTopPath` fails without this

pkg/hostpath/hostpath.go
```diff
@@ -191,15 +190,15 @@ func commonUpsertRanger(obj runtime.Object, podPath, filter string, key, value i
 	// on an update we first nuke any existing directory and then recreate it to simplify handling the case where
 	// the keys in the secret/configmap have changed such that some keys have been removed, which would translate
 	// in files having to be removed
-	if err := os.RemoveAll(podFileDir); err != nil {
+	if err := commonOSRemove(podPath); err != nil {
```
adambkaplan, line 193: nit - shouldn't the comment above be updated, since we are no longer isolating content in the type/namespace:name subdirectory?

pkg/hostpath/hostpath.go
adambkaplan: So if I understand this correctly, we can't mount shares onto existing directories in the container image?

gabemontero: We can mount onto empty directories in the container image.

So like in the entitlements demo, since `/etc/pki/entitlement` is empty when you bring up the ubi8 image, that works.

Non-empty directories and what happens, based on my testing even prior to this change, is what the driver does not take on right now.  There is a lot of tl;dr around per mount tmpfs, facilitating selinux via the kubelet, etc.  If need be let's discuss that separately.

And fwiw, we can have two shares, where
- one has a mountPath of `/data`
- and the other has a mountPath of `/data/subdir`

this worked prior to this change, and still works with this change.

pkg/hostpath/hostpath.go
gabemontero: yes it should

will update that and push in a few

pkg/hostpath/hostpath.go
gabemontero: all of which is touched upon in https://github.com/openshift/csi-driver-projected-resource#current-status-with-respect-to-the-kubernetes-volumemount-api

pkg/hostpath/hostpath.go
gabemontero: comment updated @adambkaplan

adambkaplan (approved):
/lgtm

Current limitations of mounts called out in README.

PR #35, opened 2021-05-24
BUILD-276:  add read-only volume support; miscellaneous tweaks

gabemontero:


pkg/hostpath/hostpath.go
```diff
@@ -347,6 +363,19 @@ func shareUpdateRanger(key, value interface{}) bool {
...
 				shareId, volID, oldTargetPath, err.Error())
 		}
+		if hpv.IsReadOnly() {
+			tp := hpv.GetTargetPath()
+			empty, err := isDirEmpty(tp)
+			errStr := ""
+			if err != nil {
+				errStr = err.Error()
+			}
+			klog.V(0).Infof("GGM shareUpdateRanger kubelet dir %s empty %v err %s", tp, empty, errStr)
```
gabemontero, line 373: this `hpv.IsReadOnly` if block is for https://github.com/openshift/csi-driver-projected-resource/pull/35/commits/5d7cba12da8bb202f22eee00ba7d7df69d14e482 which is a *possible* means for removing content

a) after the driver restarts
b) then permissions to the share are removed

I was able to verify this multiple times in my testing.  However, why I am able to *remove* content under the kubelet mount dir after a restart, but am unable to *update* or *change* content under the kubelet mount dir after a restart, is still a mystery to me.

Pending learning the precise *why*, I also wonder if this limited form of update support for read only volumes after driver restart will be worth the effort to maintain, document, explain, etc. moving forward will be worth it.

pkg/hostpath/hostpath.go
```diff
@@ -347,6 +363,28 @@ func shareUpdateRanger(key, value interface{}) bool {
 			klog.Warningf("share %s vol %s target path %s delete error %s",
 				shareId, volID, oldTargetPath, err.Error())
 		}
+		//TODO removing contents from a read only volume, where we employ an intermediate bind mount, is the single
+		// item in our list of update content features that still works when this driver is restarted after a Pod
+		// is started with one of our volumes.  The question is do we even bother supporting this, or just make
+		// the general statement that "results may vary" and do not claim any production level support for updating
+		// contents with read only volumes since we don't have a comprehensive solution for when the driver is restarted.
+		if hpv.IsReadOnly() {
```
gabemontero, line 371: To elaborate on the TODO comment above

I was able to verify this multiple times in my testing. However, why I am able to remove content under the kubelet mount dir after a restart, but am unable to update or change content under the kubelet mount dir after a restart, is still a mystery to me.

Pending learning the precise why, I also wonder if this limited form of update support for read only volumes after driver restart will be worth the effort to maintain, document, explain, etc. moving forward will be worth it.

examples/csi-app.yaml
```diff
@@ -11,10 +11,22 @@ spec:
...
           name: my-csi-volume
-      command: [ "sleep", "1000000" ]
+      command:
+        - sh
+        - -c
+        - |
+          while true
+          do
+            ls -la /data
+            touch /data/foo
```
gabemontero, line 21: this is (possibly short term) proof around whether the volume is read only or not ... toggling the `readOnly: true` setting below would be a way for users to play around with this element of the system.

examples/csi-app.yaml
```diff
@@ -11,10 +11,22 @@ spec:
...
+            ls -la /data
```
gabemontero, line 22: this rework is a way to prove updates to the content occur without having to `oc rsh` or `oc exec` into the pod and run `ls -la /data`

pkg/hostpath/hostpath.go
```diff
@@ -387,18 +418,34 @@ func shareUpdateRanger(key, value interface{}) bool {
...
+		if err != nil {
+			return err
+		}
+		fallthrough
+	default:
+		err := os.MkdirAll(hpv.GetVolPathAnchorDir(), 0777)
+		if err != nil {
+			return err
+		}
 	}
```
coreydaley, line 438:
```suggestion
	if hpv.IsReadOnly():
		err := os.MkdirAll(hpv.GetVolPathBindMountDir(), 0777)
		if err != nil {
			return err
		}
	}
      err := os.MkdirAll(hpv.GetVolPathAnchorDir(), 0777)
      if err != nil {
	      return err
      }

```

pkg/hostpath/hostpath.go
```diff
@@ -387,18 +418,34 @@ func shareUpdateRanger(key, value interface{}) bool {
...
 	switch strings.TrimSpace(hpv.GetSharedDataKind()) {
 	case "ConfigMap":
 		klog.V(4).Infof("mapBackingResourceToPod postlock %s configmap", hpv.GetVolID())
-		podConfigMapsPath := hpv.GetTargetPath()
+		podConfigMapsPath := ""
+		switch {
+		case readOnly:
```
coreydaley, line 444:
```suggestion
		case hpv.IsReadOnly():
```

pkg/hostpath/hostpath.go
```diff
@@ -439,7 +486,13 @@ func mapBackingResourceToPod(hpv *hostPathVolume) error {
 		objcache.RegisterConfigMapDeleteCallback(hpv.GetVolID(), deleteRangerCM)
 	case "Secret":
 		klog.V(4).Infof("mapBackingResourceToPod postlock %s secret", hpv.GetVolID())
-		podSecretsPath := hpv.GetTargetPath()
+		podSecretsPath := ""
+		switch {
+		case readOnly:
```
coreydaley, line 491:
```suggestion
		case hpv.IsReadOnly():
```

pkg/hostpath/hostpath.go
```diff
@@ -564,27 +627,33 @@ func deleteIfEmpty(name string) {
...
 
+func (hp *hostPath) innerDeleteHostpathVolume(top string) {
+	// reminder, path is filepath.Join(DataRoot, [anchor-dir | bind-dir], volID, podNamespace, podName, podUID, podSA)
+	// delete SA dir
+	err := os.RemoveAll(top)
+	if err != nil {
+		klog.Warningf("error deleting %s: %s", top, err.Error())
+	}
+	uidPath := filepath.Dir(top)
+	deleteIfEmpty(uidPath)
```
coreydaley, line 638:
```suggestion
	deleteIfEmpty(filepath.Dir(top))
```

I don't think that you need to create a variable first and then use that in the `deleteIfEmpty` function, same with the below ones

pkg/hostpath/hpv.go
```diff
@@ -111,6 +121,11 @@ func (hpv *hostPathVolume) IsAllowed() bool {
 	defer hpv.Lock.Unlock()
 	return hpv.Allowed
 }
+func (hpv *hostPathVolume) IsReadOnly() bool {
+	hpv.Lock.Lock()
+	defer hpv.Lock.Unlock()
+	return hpv.ReadOnly
```
coreydaley, line 127: So you have to lock the host path volume before you can check if it is ReadOnly? Weird.

pkg/hostpath/mount.go
```diff
@@ -0,0 +1,143 @@
...
+//   being xfs and not tmpfs
+// - with the lack of a bind option, and each pod getting its own tmpfs we have to copy the data from our emptydir
+//   based location to the targetPath here ... that is handled in hostpath.go
+type ReadWriteMany struct {
+}
+
+func (m *ReadWriteMany) makeFSMounts(anchorDir, intermediateBindMountDir, kubeletTargetDir string, mounter mount.Interface) error {
+	options := []string{}
+	if err := mounter.Mount(anchorDir, kubeletTargetDir, "tmpfs", options); err != nil {
+		var errList strings.Builder
```
coreydaley, line 56: nit: I don't think that strings.Builder is a List. Maybe a byte array?

pkg/hostpath/mount.go
```diff
@@ -0,0 +1,143 @@
...
+		return err
+	}
+	return nil
+}
+
+// WriteOnceReadMany details:
+// Our second generation version after ReadWriteMany.
+// Using an intermediate bind mount between our anchor / starting point and the kubelet's target directory, this
+// can write updates to this intermediate directory, and see those updates reflected in the Pod's volume, even though
+// it is read only for any code running code running in the Pod.
```
coreydaley, line 84:
```suggestion
// it is read only for any code running in the Pod.
```

test/framework/daemonset.go
```diff
@@ -104,8 +104,8 @@ func RestartDaemonSet(t *TestArgs) {
 	t.T.Logf("%s: csi driver pods are up with no deletion timestamps", time.Now().String())
 }
 
-//TODO presumably this can go away once we have an OLM based deploy that is also integrated with our CI
-// so that repo images built from PRs are used when setting up this driver's daemonset
+//TODO possibly this can go away once we in the OCP payload, though the csi driver operator element for all that needs to get sorted out,
```
coreydaley, line 107:
```suggestion
//TODO possibly this can go away once we are in the OCP payload, though the csi driver operator element for all that needs to get sorted out,
```

coreydaley (changes requested):
I would prefer to see if/else used where only two options are available, as switch/case seems like overkill, but like I said, that's only a preference.

pkg/hostpath/hostpath.go
```diff
@@ -223,9 +217,10 @@ func commonUpsertRanger(obj runtime.Object, podPath, filter string, key, value i
 }
 
 func commonOSRemove(dir string) error {
+	klog.V(4).Infof("commonOSRemove %s", dir)
 	// we cannot do a os.RemoveAll on the mount point, so we remove all on each file system entity
 	// off of the potential mount point
-	return filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
+	err := filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
```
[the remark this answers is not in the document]
gabemontero: I'll adjust this slightly, as I want the bracket start / stop logging ... I'll just return here, but move the klog commonOSRemove done message as a defer above

pkg/hostpath/hostpath.go
gabemontero: It is a little trickier than that, as each parameter for deleteIfEmpty builds off the last invocation, as we are working back through the file system directory tree.

That said, I'll have an alternative up that is in the spirit of what you are asking.  Let me know what you think once I post it.

pkg/hostpath/hpv.go
gabemontero: yep `go test -race` turned this up, both in the unit tests and e2e's (especially the disruptive/restart one), for all the fields of this object

essentially any concurrent activity was flagged

If you are familiar with java, I've essentially created a synchronized java bean

pkg/hostpath/mount.go
gabemontero: yeah this code is borrowed from the upstream k8s csi driver sample

strings.Builder.WriteString per its godoc appends content .... that is array / list enough for me ... I'm inclined to keep this as is

gabemontero:
ok @coreydaley updates up with a separate commit ... PTAL when you get the chance - thanks

pkg/hostpath/constants.go
```diff
@@ -1,12 +1,22 @@
...
-
 type accessType int
 
 const (
-	mountAccess accessType = iota
+	deviceID                             = "deviceID"
+	kib                       int64      = 1024
+	mib                                  = kib * 1024
+	gib                                  = mib * 1024
+	tib                                  = gib * 1024
```
coreydaley, line 10: These abbreviations read like `bits` to me instead of `bytes`. And I'm not sure where the `mib` and `tib` came from, since there are not any `i`s in them. How about something like:
```
storageCapacityKB
storageCapacityMB
storageCapacityGB
storageCapacityTB
``` 
It is a bit longer and more verbose, but also more recognizable as you read through the code.

pkg/hostpath/hostpath.go
```diff
@@ -439,7 +481,12 @@ func mapBackingResourceToPod(hpv *hostPathVolume) error {
 		objcache.RegisterConfigMapDeleteCallback(hpv.GetVolID(), deleteRangerCM)
 	case "Secret":
 		klog.V(4).Infof("mapBackingResourceToPod postlock %s secret", hpv.GetVolID())
-		podSecretsPath := hpv.GetTargetPath()
+		podSecretsPath := ""
+		if readOnly {
+			podSecretsPath = hpv.GetVolPathBindMountDir()
+		} else {
+			podSecretsPath = hpv.GetTargetPath()
+		}
```
coreydaley, line 489:
```suggestion
		podSecretsPath := hpv.GetTargetPath()
		if readOnly {
			podSecretsPath = hpv.GetVolPathBindMountDir()
		}
```

pkg/hostpath/hostpath.go
```diff
@@ -387,18 +417,30 @@ func shareUpdateRanger(key, value interface{}) bool {
...
 	switch strings.TrimSpace(hpv.GetSharedDataKind()) {
 	case "ConfigMap":
 		klog.V(4).Infof("mapBackingResourceToPod postlock %s configmap", hpv.GetVolID())
-		podConfigMapsPath := hpv.GetTargetPath()
+		podConfigMapsPath := ""
+		if readOnly {
+			podConfigMapsPath = hpv.GetVolPathBindMountDir()
+		} else {
+			podConfigMapsPath = hpv.GetTargetPath()
+		}
```
coreydaley, line 443:
```suggestion
		podConfigMapsPath := hpv.GetTargetPath()
		if readOnly {
			podConfigMapsPath = hpv.GetVolPathBindMountDir()
		}
```

pkg/hostpath/hostpath.go
```diff
@@ -564,27 +618,32 @@ func deleteIfEmpty(name string) {
...
+	}
+}
+
 // deleteVolume deletes the directory for the hostpath volume.
 func (hp *hostPath) deleteHostpathVolume(volID string) error {
 	klog.V(4).Infof("deleting hostpath volume: %s", volID)
 
-	hpv := getHPV(volID)
+	hpv := hp.getHostpathVolume(volID)
 	if hpv != nil {
```
coreydaley, line 642:
```suggestion
	if hpv := hp.getHostpathVolume(volID); hpv != nil {
```
I'm not 100% sure this one will work, but I think so?

pkg/hostpath/constants.go
gabemontero: these are more carry over from the upstream k8s repo's used as a starting point for this driver

but sure I can rename them as you suggested

pkg/hostpath/hostpath.go
gabemontero: yeah seems like it should

gabemontero:
updates pulled in / commits squashed / ptal @coreydaley thanks

putting a brief

/hold

as I want to link up @adambkaplan wrt yay or nay on https://github.com/openshift/csi-driver-projected-resource/pull/35/commits/337424c921b579386270e405c5aea363ff1312a2

pkg/hostpath/hostpath.go
adambkaplan: I'm fine leaving this for now - Tech Preview means we have freedom to explore these things and figure out what we can ultimately support.

gabemontero: sounds good thanks

PR #36, opened 2021-05-25
BUILD-261: update install.md now that releases page is set up

gabemontero:
/assign @adambkaplan

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -108,13 +108,11 @@ daemonset.apps/csi-hostpathplugin created
 
 ## Installing from the release page  (only developer preview level support)
 
-1. Run the following command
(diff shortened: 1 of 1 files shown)


adambkaplan:
#37 is an alternative - a few more changes there.

docs/install.md
```diff
@@ -108,13 +108,11 @@ daemonset.apps/csi-hostpathplugin created
 
 ## Installing from the release page  (only developer preview level support)
 
-1. Run the following command
+Follow the instructions for any of the releases listed on [the Releases page](https://github.com/openshift/csi-driver-projected-resource/releases) that you are interested in, starting with release 'v0.4.8-rc.0'
```
adambkaplan, line 111: I don't like redirecting folks to another page in docs. In #37 I let the user pick the version via an env var.

gabemontero: how do they know which value to set to the env var unless they go to the other page?

gabemontero:
I've posted comments in #37 and asked for retitle ... we'll go with that one once my input is addressed.

PR #39, opened 2021-07-28
Rename api group to `sharedresource.openshift.io`

adambkaplan:
- Rename api group to sharedresource.storage.openshift.io.
- Update base CRD definition and type registration with new group name.
- Regenerated deployment YAMLs.
- Rename pkg/api/projectedresource to pkg/api/sharedresource
- Rename arguments for update/generated.sh to use new name
- Regenerate informers and listers
- Rename golang package imports
- Update controller with new informer factory name
- Update e2e tests to use new client
- Rename CRD manifest to "sharedresource"
- Move base YAML manifest to the sharedresource/v1alpha1 directory
- Add rename to update-crd.sh


coreydaley:
@adambkaplan I think that https://github.com/openshift/csi-driver-shared-resource/blob/master/Dockerfile needs to be updated to reflect the new github repository name and go package name?

coreydaley:
```
2021-08-03T16:13:07.904133683Z STEP 8: COPY --from=builder /go/src/github.com/openshift/csi-driver-projected-resource/_output/csi-driver-projected-resource /usr/bin/
2021-08-03T16:13:08.690272181Z STEP 9: ENTRYPOINT null
2021-08-03T16:13:08.690272181Z STEP 10: CMD ["/usr/bin/csi-driver-projected-resource"]
```

adambkaplan:
I honestly don't know how CI works without registering the `CSIDriver` object. Kube can't create pods if the CSI driver isn't registered with the cluster. I don't see anything in openshift/release or the make commands which indicate we register the CSI driver object in our test.

coreydaley:
@adambkaplan 
```
 "message": "MountVolume.SetUp failed for volume \"my-csi-volume\" : rpc error: code = InvalidArgument desc = the csi driver volumeAttribute 'share' reference had an error: share.projectedresource.storage.openshift.io \"e2e-test-csi-driver-projected-resource-kdjtt\" not found",
```

adambkaplan:
@coreydaley this is the moment where I screamed at the top of my lungs, "WHAT'S GOING ON?!" [1]

I suspect that either:

1. We aren't deploying the driver we want in CI
2. The client code that we're generating in this tree is somehow hard-coding the previous apigroup.

The only way I think we can address the situation is close this PR, and start fresh with generated code from openshift/api and openshift/client-go vendored in.

[1] https://www.youtube.com/watch?v=6NXnxTNIWkc

PR #46, opened 2021-08-17
BUILD-264: Refresh-Resources Mode and Configurable Namespaces

otaviof:
This pull-request adds two more flags on the operator (controller) part of the CSI Driver. And modifies the test infrastructure to be able to deploy with different arguments.

### Flags

Using `--refreshresources=false` (default's `true`) flag the operator will only watch `Share` resources, and will read the actual backing resource object just before mounting the volume. Previously it use to watch over most of the `ConfigMap` and `Secret` resources. By using this flag the operator will consume less resources, and won't watch for eventual resources modifications.

Additionally, the list of ignored namespaces can be edited using `--ignorenamespace`. The original list of ignored namespaces is kept by default.

### E2E

A new target is added, `test-e2e-no-refreshresources` which deploys the operator using this mode and runs the configured test-suite against. For instance:

```bash
make test-e2e-no-refreshresources DRIVER_IMAGE="..." NODE_REGISTRAR_IMAGE="..."
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,21 @@ REPOSITORY ?= openshift
 # Allows overriding of the tag to generate for the image
 TAG ?= latest
 
+# allows overwrite of node registrar image
+NODE_REGISTRAR_IMAGE ?=
+# allows overwite of CSI driver image
+DRIVER_IMAGE ?=
+# when non-empty it enables "--refreshresource=false" deployment
+DEPLOY_MODE ?=
+
+# amount of pods the daemonset will produce before starting end-to-end tests
+DAEMONSET_PODS ?= 3
+
+# end-to-end test suite name
+TEST_SUITE ?= normal
+# ent-to-end test timeout
+TEST_TIMEOUT ?= 30m
+
 LDFLAGS ?= '-extldflags "-static"'
 
 .DEFAULT_GOAL := help
@@ -26,27 +41,30 @@ test: ## Run unit tests. Example: make test
 .PHONY: test
 
 deploy:
-	# For local testing set DRIVER_IMAGE to whatever image you produced via 'make build-image'
-	./deploy/deploy.sh
+	NODE_REGISTRAR_IMAGE=$(NODE_REGISTRAR_IMAGE) DRIVER_IMAGE=$(DRIVER_IMAGE) ./deploy/deploy.sh $(DEPLOY_MODE)
 .PHONY: deploy
 
-TEST_SUITE ?= normal
-TEST_TIMEOUT ?= 30m
-test-e2e-no-deploy:
+# overwrites the deployment mode variable to disable refresh-resources
+deploy-no-refreshresources: DEPLOY_MODE = "no-refreshresources"
+deploy-no-refreshresources: deploy
+
+crd:
 	# temporary creation of CRD until it lands in openshift/api, openshift/openshift-apiserver, etc.
 	oc apply -f ./deploy/0000_10_projectedresource.crd.yaml
-	TEST_SUITE=$(TEST_SUITE) TEST_TIMEOUT=$(TEST_TIMEOUT) ./hack/test-e2e.sh
 
-test-e2e: deploy test-e2e-no-deploy
-.PHONY: test-e2e
+test-e2e-no-deploy: crd
+	TEST_SUITE=$(TEST_SUITE) TEST_TIMEOUT=$(TEST_TIMEOUT) DAEMONSET_PODS=$(DAEMONSET_PODS) ./hack/test-e2e.sh
+.PHONY: test-e2e-no-deploy
 
-test-e2e-slow: deploy
-	TEST_SUITE="slow" ./hack/test-e2e.sh
-.PHONY: test-e2e
+test-e2e: crd deploy test-e2e-no-deploy
 
-test-e2e-disruptive: deploy
-	TEST_SUITE="disruptive" ./hack/test-e2e.sh
-.PHONY: test-e2e
+test-e2e-no-refreshresources: crd deploy-no-refreshresources test-e2e-no-deploy
+
+test-e2e-slow: TEST_SUITE = "slow"
+test-e2e-slow: deploy test-e2e
+
+test-e2e-disruptive: TEST_SUITE = "disruptive" 
+test-e2e-disruptive: deploy test-e2e
 
 verify: ## Run verifications. Example: make verify
 	go vet ./cmd/... ./pkg/... ./test/...
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -11,9 +11,11 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
 
 	v1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/csi-driver-shared-resource/pkg/client"
 	"github.com/openshift/csi-driver-shared-resource/pkg/controller"
 	"github.com/openshift/csi-driver-shared-resource/pkg/hostpath"
 )
@@ -26,6 +28,8 @@ var (
 	maxVolumesPerNode   int64
 	version             string
 	shareRelistInterval string
+	refreshResources    bool
+	ignoredNamespaces   []string
 
 	shutdownSignals      = []os.Signal{os.Interrupt, syscall.SIGTERM}
 	onlyOneSignalHandler = make(chan struct{})
@@ -37,7 +41,19 @@ var rootCmd = &cobra.Command{
 	Short:   "",
 	Long:    ``,
 	Run: func(cmd *cobra.Command, args []string) {
-		driver, err := hostpath.NewHostPathDriver(hostpath.DataRoot, hostpath.VolumeMapRoot, driverName, nodeID, endPoint, maxVolumesPerNode, version)
+		var kubeClient kubernetes.Interface
+		var err error
+
+		if !refreshResources {
+			fmt.Println("Refresh-Resources disabled, loading a Kubernetes client for HostPathDriver")
+
+			if kubeClient, err = loadKubernetesClientset(); err != nil {
+				fmt.Printf("Failed to load Kubernetes API client: %s", err.Error())
+				os.Exit(1)
+			}
+		}
+
+		driver, err := hostpath.NewHostPathDriver(hostpath.DataRoot, hostpath.VolumeMapRoot, driverName, nodeID, endPoint, maxVolumesPerNode, version, kubeClient)
 		if err != nil {
 			fmt.Printf("Failed to initialize driver: %s", err.Error())
 			os.Exit(1)
@@ -68,6 +84,17 @@ func init() {
 	rootCmd.Flags().Int64Var(&maxVolumesPerNode, "maxvolumespernode", 0, "limit of volumes per node")
 	rootCmd.Flags().StringVar(&shareRelistInterval, "share-relist-interval", "",
 		"the time between controller relist on the share resource expressed with golang time.Duration syntax(default=10m")
+	rootCmd.Flags().BoolVar(&refreshResources, "refreshresources", true, "watch for resource updates")
+	rootCmd.Flags().StringSliceVar(&ignoredNamespaces, "ignorenamespace", []string{}, "Specify a namespace to be ignored by the controller")
+}
+
+// loadKubernetesClientset instantiate a clientset using local config.
+func loadKubernetesClientset() (kubernetes.Interface, error) {
+	kubeRestConfig, err := client.GetConfig()
+	if err != nil {
+		return nil, err
+	}
+	return kubernetes.NewForConfig(kubeRestConfig)
 }
 
 func runOperator() {
@@ -81,7 +108,7 @@ func runOperator() {
 			shareRelist = controller.DefaultResyncDuration
 		}
 	}
-	c, err := controller.NewController(shareRelist)
+	c, err := controller.NewController(shareRelist, refreshResources, ignoredNamespaces)
 	if err != nil {
 		fmt.Printf("Failed to set up controller: %s", err.Error())
 		os.Exit(1)
--- a/deploy/csi-hostpath-plugin.yaml
+++ b/deploy/csi-hostpath-plugin.yaml
@@ -70,6 +70,30 @@ spec:
             - "--v=4"
             - "--endpoint=$(CSI_ENDPOINT)"
             - "--nodeid=$(KUBE_NODE_NAME)"
+            - --ignorenamespace=openshift-machine-api
+            - --ignorenamespace=openshift-kube-apiserver
+            - --ignorenamespace=openshift-kube-apiserver-operator
+            - --ignorenamespace=openshift-kube-scheduler
+            - --ignorenamespace=openshift-kube-controller-manager
+            - --ignorenamespace=openshift-kube-controller-manager-operator
+            - --ignorenamespace=openshift-kube-scheduler-operator
+            - --ignorenamespace=openshift-console-operator
+            - --ignorenamespace=openshift-controller-manager
+            - --ignorenamespace=openshift-controller-manager-operator
+            - --ignorenamespace=openshift-cloud-credential-operator
+            - --ignorenamespace=openshift-authentication-operator
+            - --ignorenamespace=openshift-service-ca
+            - --ignorenamespace=openshift-kube-storage-version-migrator-operator
+            - --ignorenamespace=openshift-config-operator
+            - --ignorenamespace=openshift-etcd-operator
+            - --ignorenamespace=openshift-apiserver-operator
+            - --ignorenamespace=openshift-cluster-csi-drivers
+            - --ignorenamespace=openshift-cluster-storage-operator
+            - --ignorenamespace=openshift-cluster-version
+            - --ignorenamespace=openshift-image-registry
+            - --ignorenamespace=openshift-machine-config-operator
+            - --ignorenamespace=openshift-sdn
+            - --ignorenamespace=openshift-service-ca-operator
           env:
             - name: CSI_ENDPOINT
               value: unix:///csi/csi.sock
--- a/deploy/deploy.sh
+++ b/deploy/deploy.sh
@@ -16,54 +16,122 @@
 set -e
 set -o pipefail
 
+# when not empty it will run CSI driver with "--refreshresources=false" flag.
+NO_REFRESH_RESOURCES="${1}"
+
+# customize images used by registrar and csi-driver containers
+NODE_REGISTRAR_IMAGE="${NODE_REGISTRAR_IMAGE:-}"
+DRIVER_IMAGE="${DRIVER_IMAGE:-}"
+
 # BASE_DIR will default to deploy
 BASE_DIR="deploy"
 DEPLOY_DIR="_output/deploy"
+
+KUSTOMIZATION_FILE="${DEPLOY_DIR}/kustomization.yaml"
 
 function run () {
     echo "$@" >&2
     "$@"
 }
 
-echo "Creating deploy directory"
-rm -rf "${DEPLOY_DIR}"
+# initialize a kustomization.yaml file, listing all other yaml files as resources.
+function kustomize_init () {
+    cat <<EOS > ${KUSTOMIZATION_FILE}
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+EOS
+
+  for f in $(find "${DEPLOY_DIR}"/*.yaml |grep -v kustomization |sort); do
+    f=$(basename ${f})
+    echo "## ${f}"
+    echo "  - ${f}" >> ${KUSTOMIZATION_FILE}
+  done
+
+  echo "images:" >> ${KUSTOMIZATION_FILE}
+}
+
+# creates a new entry with informed name and target image. The target image is split on URL and tag.
+function kustomize_set_image () {
+  local NAME=${1}
+  local TARGET=${2}
+
+  # splitting target image in URL and tag
+  IFS=':' read -a PARTS <<< ${TARGET}
+
+  cat <<EOS >> ${KUSTOMIZATION_FILE}
+  - name: ${NAME}
+    newName: ${PARTS[0]}
+    newTag: ${PARTS[1]}
+EOS
+}
+
+# patches the CSI DaemonSet primary container to include one more argument.
+function kustomize_add_arg () {
+  local ARG=${1}
+  local FILE="args-patch-${ARG}.json"
+  cat <<EOS > ${DEPLOY_DIR}/${FILE}
+[
+  {
+    "op": "add",
+    "path": "/spec/template/spec/containers/1/args/-",
+    "value": "${ARG}"
+  }
+]
+EOS
+  cat <<EOS >> ${KUSTOMIZATION_FILE}
+patchesJson6902:
+  - path: ${FILE}
+    target:
+      group: apps
+      version: v1
+      kind: DaemonSet
+      namespace: openshift-cluster-csi-drivers
+      name: csi-hostpathplugin
+EOS
+}
+
+# uses `oc wait` to wait for CSI driver pod to reach condition ready.
+function wait_for_pod () {
+  oc --namespace="openshift-cluster-csi-drivers" wait pod \
+    --for="condition=Ready=true" \
+    --selector="app=csi-hostpathplugin" \
+    --timeout="5m"
+}
+
+echo "# Creating deploy directory at '${DEPLOY_DIR}'"
+
+rm -rf "${DEPLOY_DIR}" || true
 mkdir -p "${DEPLOY_DIR}"
 
-defaultRegistrarImage="quay.io/openshift/origin-csi-node-driver-registrar:4.9.0"
-registrarImage=${NODE_REGISTRAR_IMAGE:-${defaultRegistrarImage}}
-defaultDriverImage="quay.io/openshift/origin-csi-driver-shared-resource:4.9.0"
-driverImage=${DRIVER_IMAGE:-${defaultDriverImage}}
+cp -r -v "${BASE_DIR}"/* "${DEPLOY_DIR}"
 
-cp -r "${BASE_DIR}"/* "${DEPLOY_DIR}"
+echo "# Customizing resources..."
 
-echo "Deploying using node driver registrar ${registrarImage}"
-echo "Deploying using csi driver ${driverImage}"
+# initializing kustomize and adding the all resource files it should use
+kustomize_init
 
-# Replace the image refs in the canonical deployment YAML with env var overrides
-sed -i -e "s|${defaultRegistrarImage}|${registrarImage}|g" \
-  -e "s|${defaultDriverImage}|${driverImage}|g" \
-  "${DEPLOY_DIR}/csi-hostpath-plugin.yaml"
+if [ ! -z "${NODE_REGISTRAR_IMAGE}" ] ; then
+  echo "# Patching node-registrar image to use '${NODE_REGISTRAR_IMAGE}'"
+  kustomize_set_image "quay.io/openshift/origin-csi-node-driver-registrar" "${NODE_REGISTRAR_IMAGE}"
+fi
+
+if [ ! -z "${DRIVER_IMAGE}" ] ; then
+  echo "# Patching node-csi-driver image to use '${DRIVER_IMAGE}'"
+  kustomize_set_image "quay.io/openshift/origin-csi-driver-shared-resource" "${DRIVER_IMAGE}"
+fi
+
+# adding to disable refresh-resources using kustomize-v3 approach (embedded on `oc`)
+if [ ! -z "${NO_REFRESH_RESOURCES}" ] ; then
+  echo "# Patching DaemonSet container to use '--refreshresources=false' flag"
+  kustomize_add_arg "--refreshresources=false"
+fi
 
 # deploy hostpath plugin and registrar sidecar
-echo "deploying csi driver components"
-for i in $(find "${DEPLOY_DIR}"/*.yaml | sort); do
-    echo "   $i"
-    run oc apply -f "$i"
-done
+echo "# Deploying csi driver components"
+run oc apply --kustomize ${DEPLOY_DIR}/
 
-# Wait until all pods are running.
-expected_running_pods=3
-cnt=0
-while [ "$(oc get pods -n openshift-cluster-csi-drivers 2>/dev/null | grep -c '^csi-hostpath.* Running ')" -lt ${expected_running_pods} ]; do
-    if [ $cnt -gt 30 ]; then
-        echo "$(oc get pods 2>/dev/null | grep -c '^csi-hostpath.* Running ') running pods:"
-        oc describe pods
-
-        echo >&2 "ERROR: hostpath deployment not ready after over 5min"
-        exit 1
-    fi
-    echo "$(date +%H:%M:%S) waiting for hostpath deployment to complete, attempt #$cnt"
-    cnt=$((cnt + 1))
-    sleep 10
-done
-
+# waiting for all pods to reach condition ready
+echo "# Waiting for pods to be ready..."
+wait_for_pod || sleep 15 && wait_for_pod
--- a/docs/content-update-details.md
+++ b/docs/content-update-details.md
@@ -2,7 +2,7 @@
 
 ### Excluded OCP namespaces
 
-The current list of namespaces excluded from the controller's watches:
+The current list of namespaces excluded by default from the controller's watches:
 
 - kube-system
 - openshift-machine-api
@@ -30,7 +30,14 @@ The current list of namespaces excluded from the controller's watches:
 - openshift-sdn
 - openshift-service-ca-operator
 
-The list is not yet configurable, but most likely will become so as the project's lifecycle progresses.
+The list is configurable by informing `--ignorenamespace` to the `hostpath` plugin instance. The
+plugin can also be configured with `--refreshresources` flag, which makes the controller keep a warm
+cache of `ConfigMap` and `Secret` objects, and as they change the controller will follow the updates.
+
+When `--refreshresources` is disabled (i.e. `--refreshresources=false`), the controller will read the
+backing-resource (`.spec.backingResource`) just before mounting the volume instead of keeping a warm
+cache. Additionally, every volume mount will be always read-only preventing tampering with the data
+provided by this CSI driver.
 
 Allowing the disabling processing of updates, or switching the default for the system as not dealing with
 updates, but then allowing for opting into updates, is also under consideration.
--- a/pkg/cache/common.go
+++ b/pkg/cache/common.go
@@ -2,6 +2,7 @@ package cache
 
 import (
 	"fmt"
+	"strings"
 	"sync"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -17,6 +18,15 @@ func GetKey(o interface{}) string {
 
 func BuildKey(namespace, name string) string {
 	return namespace + ":" + name
+}
+
+// SplitKey splits the shared-data-key into namespace and name.
+func SplitKey(key string) (string, string, error) {
+	s := strings.Split(key, ":")
+	if len(s) != 2 || s[0] == "" || s[1] == "" {
+		return "", "", fmt.Errorf("unable to split key '%s' into namespace and name", key)
+	}
+	return s[0], s[1], nil
 }
 
 func buildCallbackMap(key, value interface{}) *sync.Map {
--- a/pkg/cache/configmaps.go
+++ b/pkg/cache/configmaps.go
@@ -1,9 +1,12 @@
 package cache
 
 import (
+	"context"
 	"sync"
 
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
 
 	sharev1alpha1 "github.com/openshift/csi-driver-shared-resource/pkg/api/sharedresource/v1alpha1"
@@ -24,7 +27,7 @@ On the use of sync.Map, see the comments in share.go
 
 var (
 	// configmaps is our global configmap id (namespace + name) to configmap map, where entries are populated from
-	// contorller events; it serves to facilitate quick lookup during share event processing, when the share references
+	// controller events; it serves to facilitate quick lookup during share event processing, when the share references
 	// a configmap
 	configmaps = sync.Map{}
 	// configmapUpsertCallbacks has a key of the CSI volume ID and a value of the function to be called when a given
@@ -54,6 +57,23 @@ func GetConfigMap(key interface{}) *corev1.ConfigMap {
 		cm, _ := obj.(*corev1.ConfigMap)
 		return cm
 	}
+	return nil
+}
+
+// SetConfigMap based on the shared-data-key, which contains the resource's namespace and name, this
+// method can fetch and store it on cache.
+func SetConfigMap(kubeClient kubernetes.Interface, sharedDataKey string) error {
+	ns, name, err := SplitKey(sharedDataKey)
+	if err != nil {
+		return err
+	}
+
+	cm, err := kubeClient.CoreV1().ConfigMaps(ns).Get(context.TODO(), name, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	UpsertConfigMap(cm)
 	return nil
 }
 
--- a/pkg/cache/secrets.go
+++ b/pkg/cache/secrets.go
@@ -1,8 +1,11 @@
 package cache
 
 import (
+	"context"
 	"sync"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
 
 	sharev1alpha1 "github.com/openshift/csi-driver-shared-resource/pkg/api/sharedresource/v1alpha1"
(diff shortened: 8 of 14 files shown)


pkg/controller/controller.go
```diff
@@ -112,46 +118,66 @@ func NewController(shareRelist time.Duration) (*Controller, error) {
...
-		secInformer:          informerFactory.Core().V1().Secrets().Informer(),
 		shareInformer:        shareInformerFactory.Sharedresource().V1alpha1().Shares().Informer(),
 		listers:              client.GetListers(),
 	}
 
-	client.SetConfigMapsLister(c.informerFactory.Core().V1().ConfigMaps().Lister())
-	client.SetSecretsLister(c.informerFactory.Core().V1().Secrets().Lister())
+	if !readOnly {
+		c.cfgMapWorkqueue = workqueue.NewNamedRateLimitingQueue(
+			workqueue.DefaultControllerRateLimiter(), "projected-resource-configmap-changes")
```
gabemontero, line 134: projected -> shared

pkg/controller/controller.go
```diff
@@ -112,46 +118,66 @@ func NewController(shareRelist time.Duration) (*Controller, error) {
...
+		c.secretWorkqueue = workqueue.NewNamedRateLimitingQueue(
+			workqueue.DefaultControllerRateLimiter(), "projected-resource-secret-changes")
```
gabemontero, line 136: projected -> shared

pkg/hostpath/hostpath.go
```diff
@@ -524,8 +528,27 @@ func mapBackingResourceToPod(hpv *hostPathVolume) error {
 	return nil
 }
 
+func (hp *hostPath) populateCache(hpv *hostPathVolume) error {
+	klog.V(4).Infof("populating object-cache before creating mount scaffold")
+	switch strings.TrimSpace(hpv.GetSharedDataKind()) {
+	case "ConfigMap":
+		return objcache.SetConfigMap(hp.kubeClient, hpv.GetSharedDataKey())
+	case "Secret":
```
gabemontero, line 536: I am assuming something similar to what is done above for ConfigMaps is coming and will be here once WIP is removed

pkg/hostpath/hpv.go
```diff
@@ -76,6 +76,18 @@ func (hpv *hostPathVolume) GetTargetPath() string {
 	defer hpv.Lock.Unlock()
 	return hpv.TargetPath
 }
+
+// TODO: document!
```
gabemontero, line 80: The need to document aside, this method feels correct in that we don't want/need that extra layer of indirection in our file system mount stack when we are read only

gabemontero (commented):
Generally looks like I would have expected @otaviof and in the spirit of our 1-1 kick off meeting / code walk through a while back... cool deal !!

A minor naming suggestion or two, otherwise just some in line observations I suspect are already on the path of resolution given you still have the WIP tag on this PR.

And for out of band observations, again with the qualifier that this is still WIP, is that as I'm sure you are already on path for, the unit/e2e stuff will need to be tweaked based on read only vs. read write.

But given your changes so far did not impact the e2e's run are certainly encouraging.

And I am not concerned that a few unit tests failed / need tweaks.  Again, not surprising for a WIP PR.

gabemontero:
FYI ... (re)confirmation that the CI changes from @adambkaplan and I are pulling in the image from this driver built from this PR's code.

From the ci/prow/images job:

```
2021-08-18T07:15:27.472282690Z Pushing image image-registry.openshift-image-registry.svc:5000/ci-op-4t9mg8y6/pipeline:csi-driver-shared-resource ...
2021-08-18T07:15:27.472944655Z Getting image source signatures
2021-08-18T07:15:27.586030731Z Copying blob sha256:983f1c02b066c17f8c8188228a50cc64976feddb94f2562257ea8f4c9475fe67
2021-08-18T07:15:27.588047974Z Copying blob sha256:296e14ee24149e14d573a1fbf5b5a625c7bb0cc22f5b2a8b180e833258187948
2021-08-18T07:15:27.592182814Z Copying blob sha256:356f18f3a935b2f226093720b65383048249413ed99da72c87d5be58cc46661c
2021-08-18T07:15:27.596988152Z Copying blob sha256:dc1eaa2cd0b85b00e4014fb863457a55b18d741eb2062dfcd3b65ad34c0cf242
2021-08-18T07:15:27.601363055Z Copying blob sha256:7e70cd65b8ce38214c95d786a377472d5c215c43735b8911c5c00c7fdb3bc96e
2021-08-18T07:15:28.914059592Z Copying blob sha256:d14dfac288a514336f4b2caedc4e3a3bdb9ed1ca0b130161084ba026452b8f0c
2021-08-18T07:15:30.437787840Z Copying config sha256:ae369772b70875fefe08f6d64cad9a563f1985a15f896a639e5849a7e5c5d542
2021-08-18T07:15:32.690909003Z Writing manifest to image destination
2021-08-18T07:15:33.009809004Z Storing signatures
2021-08-18T07:15:33.009875104Z Successfully pushed image-registry.openshift-image-registry.svc:5000/ci-op-4t9mg8y6/pipeline@sha256:a0acd20ddff72e106dcab0e7e87ae767d4bf3b2e4b1b12dd7e768c680cdc2929
2021-08-18T07:15:33.045910668Z Push successful
```

from one of the test e2e's:

`Deploying using csi driver registry.build02.ci.openshift.org/ci-op-4t9mg8y6/pipeline@sha256:a0acd20ddff72e106dcab0e7e87ae767d4bf3b2e4b1b12dd7e768c680cdc2929`

same SHA for the "pipeline" imagestream

There is an assumption in my statement above that https://github.com/openshift/csi-driver-shared-resource/blob/master/deploy/deploy.sh#L42-L45 worked.

Certainly @otaviof as you move forward with this you are not entirely convinced, I would suggest starting in that deploy.sh file and cat'ing the contents of that csi-hostpath-plugin.yaml, along with temporary dumps of the csi driver pod logs for any new tracing you added.

pkg/controller/controller.go
otaviof: Thanks! Fixed on the last push.

pkg/hostpath/hostpath.go
otaviof: Indeed. Please consider the latest changes, it encompasses Secrets as well.

cmd/main.go
```diff
@@ -68,6 +83,15 @@ func init() {
 	rootCmd.Flags().Int64Var(&maxVolumesPerNode, "maxvolumespernode", 0, "limit of volumes per node")
 	rootCmd.Flags().StringVar(&shareRelistInterval, "share-relist-interval", "",
 		"the time between controller relist on the share resource expressed with golang time.Duration syntax(default=10m")
+	rootCmd.Flags().BoolVar(&readOnly, "readonly", false, "does not watch for resource updates")
```
coreydaley, line 86: `readonly` does not make sense to me here with the `does not watch for resource updates` description. What is it setting as `readonly`? Maybe it should be something like `no-watch`?

cmd/main.go
```diff
@@ -37,7 +40,19 @@ var rootCmd = &cobra.Command{
...
 	Long:    ``,
 	Run: func(cmd *cobra.Command, args []string) {
-		driver, err := hostpath.NewHostPathDriver(hostpath.DataRoot, hostpath.VolumeMapRoot, driverName, nodeID, endPoint, maxVolumesPerNode, version)
+		var kubeClient kubernetes.Interface
+		var err error
+
+		if readOnly {
+			fmt.Println("Read-Only mode: Loading Kubernetes client for HostPathDriver")
+
+			if kubeClient, err = loadKubeConfig(); err != nil {
```
coreydaley, line 49: Aren't you basically just duplicating here what the `loadKubeConfig` function is doing? checking for an error etc. It would probably be cleaner to just drop the `loadKubeConfig` function and change this to
`if kubeClient, err = client.GetConfig(); err != nil {`

cmd/main.go
```diff
@@ -68,6 +83,15 @@ func init() {
...
+	rootCmd.Flags().BoolVar(&readOnly, "readonly", false, "does not watch for resource updates")
+}
+
+func loadKubeConfig() (*kubernetes.Clientset, error) {
+	kubeRestConfig, err := client.GetConfig()
+	if err != nil {
+		return nil, err
+	}
+	return kubernetes.NewForConfig(kubeRestConfig)
 }
```
coreydaley, line 95: Drop this function in favor of just using `client.GetConfig` directly.

pkg/cache/common.go
```diff
@@ -19,6 +20,15 @@ func BuildKey(namespace, name string) string {
 	return namespace + ":" + name
 }
 
+// SplitKey splits the shared-data-key in namespace and name.
```
coreydaley, line 23: s/in/into/

pkg/cache/common.go
```diff
@@ -19,6 +20,15 @@ func BuildKey(namespace, name string) string {
...
+func SplitKey(key string) (string, string, error) {
+	s := strings.Split(key, ":")
+	if len(s) != 2 || s[0] == "" || s[1] == "" {
+		return "", "", fmt.Errorf("unable to split key '%s' in namespace and name", key)
```
coreydaley, line 27: s/in/into/

pkg/cache/configmaps.go
```diff
@@ -57,6 +60,24 @@ func GetConfigMap(key interface{}) *corev1.ConfigMap {
...
+// SetConfigMap based on the shared-data-key, which contains the resource's namespace and name, this
+// method can fetch and store it on cache.
+func SetConfigMap(kubeClient kubernetes.Interface, sharedDataKey string) error {
+	ns, name, err := SplitKey(sharedDataKey)
+	if err != nil {
+		return err
+	}
+
+	ctx := context.TODO()
+	cm, err := kubeClient.CoreV1().ConfigMaps(ns).Get(ctx, name, metav1.GetOptions{})
```
coreydaley, line 72: If you aren't going to use the context anywhere else, why not pass the `context.TODO()` directly?
`cm, err := kubeClient.CoreV1().ConfigMaps(ns).Get(context.TODO(), name, metav1.GetOptions{})`

pkg/cache/secrets.go
```diff
@@ -58,6 +61,24 @@ func GetSecret(key interface{}) *corev1.Secret {
...
+// SetSecret based on the shared-data-key, which contains the resource's namespace and name, this
+// method can fetch and store it on cache.
+func SetSecret(kubeClient kubernetes.Interface, sharedDataKey string) error {
+	ns, name, err := SplitKey(sharedDataKey)
+	if err != nil {
+		return err
+	}
+
+	ctx := context.TODO()
+	secret, err := kubeClient.CoreV1().Secrets(ns).Get(ctx, name, metav1.GetOptions{})
```
coreydaley, line 73: Same thing here
`secret, err := kubeClient.CoreV1().Secrets(ns).Get(context.TODO(), name, metav1.GetOptions{})`

coreydaley (changes requested):
Overall I don't think that `readonly` make sense as the flag name, maybe `readonce` would be better?

pkg/hostpath/hpv.go
otaviof: That snippet was a left over, removed.

cmd/main.go
otaviof: The method is poorly named, it should reflect that in fact it loads the Kubernetes Config and then instantiate a `clientset`. Both actions can produce error, therefore the method holds them together. I'll rename it :-)

cmd/main.go
otaviof: Please consider previous [comment](https://github.com/openshift/csi-driver-shared-resource/pull/46#discussion_r694669253).

pkg/cache/common.go
otaviof: Thanks!

pkg/cache/common.go
otaviof: Thanks :-)

pkg/cache/configmaps.go
otaviof: Indeed, please consider latest changes.

pkg/cache/secrets.go
otaviof: Sure, please consider latest changes.

cmd/main.go
otaviof: I agree, the name `read-only` doesn't resemble what the feature does. However, `no-watch` also doesn't either, given with this flag enabled it will only watch for `Share` resources, no longer for `Secret` and `ConfigMap`.

What other names would we suggest for this flag?

gabemontero:
@otaviof - as part of confirming that your `kustomize` related changes are doing what we want, please examine the `ci/prow/images` job log to locate the imagestreamtag image reference for the image built from this PR's code changes, and then go to one of the e2e job run logs on prow, and show where the deploy scripts and test output confirms that the PR test image is being used for the driver's `DaemonSet`.

I did a similar exercise back in https://github.com/openshift/csi-driver-shared-resource/pull/46#issuecomment-901356817

Now making another prelim pass at the code changes, still trying to take into account that you have the WIP marker on this PR and you are presumably not done making changes.

Thanks.

cmd/main.go
gabemontero: perhaps to verbose, but `no-watch-since-read-only` ?

cmd/main.go
```diff
@@ -68,6 +84,17 @@ func init() {
 	rootCmd.Flags().Int64Var(&maxVolumesPerNode, "maxvolumespernode", 0, "limit of volumes per node")
 	rootCmd.Flags().StringVar(&shareRelistInterval, "share-relist-interval", "",
 		"the time between controller relist on the share resource expressed with golang time.Duration syntax(default=10m")
+	rootCmd.Flags().BoolVar(&readOnly, "readonly", false, "does not watch for resource updates")
+	rootCmd.Flags().StringSliceVar(&ignoredNamespaces, "ignorenamespace", []string{}, "namespaces ignored by the controller")
```
gabemontero, line 88: I wonder if in addition to the command line arg, we document and employ a specific config map in the openshift-cluster-csi-drivers namespace that our driver uses for analogous config for his stuff, we associated keys in the config map for each of these params

we would also establish an order of precedence between command line arg and config map

gabemontero: continue to have our default, non "system" OCP namespaces for now, though over time, we will probably have to adjust this as the rotating cert scenario withing OCP starts leveraging our stuff

pkg/controller/controller.go
```diff
@@ -112,46 +89,65 @@ func NewController(shareRelist time.Duration) (*Controller, error) {
...
 	}
 
-	client.SetConfigMapsLister(c.informerFactory.Core().V1().ConfigMaps().Lister())
-	client.SetSecretsLister(c.informerFactory.Core().V1().Secrets().Lister())
+	if !readOnly {
+		c.cfgMapWorkqueue = workqueue.NewNamedRateLimitingQueue(
+			workqueue.DefaultControllerRateLimiter(), "shared-resource-configmap-changes")
+		c.secretWorkqueue = workqueue.NewNamedRateLimitingQueue(
+			workqueue.DefaultControllerRateLimiter(), "shared-resource-secret-changes")
+		c.cfgMapInformer = informerFactory.Core().V1().ConfigMaps().Informer()
```
gabemontero, line 107: if we do employ an ocp csi driver namespace for our config, we would want to always watch that one specifically, even in read only mode

gabemontero (commented):
Added some prelim, the PR is still WIP, comments @otaviof - thanks

otaviof:
> @otaviof - as part of confirming that your `kustomize` related changes are doing what we want, please examine the `ci/prow/images` job log to locate the imagestreamtag image reference for the image built from this PR's code changes, and then go to one of the e2e job run logs on prow, and show where the deploy scripts and test output confirms that the PR test image is being used for the driver's `DaemonSet`.
> 
> I did a similar exercise back in [#46 (comment)](https://github.com/openshift/csi-driver-shared-resource/pull/46#issuecomment-901356817)

Okay, given the changes in `deploy.sh` and the introduction of `kustomize` is only fair to demostrante the new scripting is performing as expected.

As for the image replacement we use to tackle with `sed`, those are now working as expected by inspecting the `deploy.sh` [script output][deployScriptOutput], i.e.:

```sh
$ curl --silent https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_csi-driver-shared-resource/46/pull-ci-openshift-csi-driver-shared-resource-master-e2e-aws-csi-driver/1430137437895528448/artifacts/e2e-aws-csi-driver/test/build-log.txt |grep -i 'patching.*image'

# Patching node-registrar image to use 'registry.build01.ci.openshift.org/ci-op-tv0f2inb/stable@sha256:b8e2d637f14e04658595393df59d99ac12284402c64c18c0a0aeecb29c70b1fb'
# Patching node-csi-driver image to use 'registry.build01.ci.openshift.org/ci-op-tv0f2inb/pipeline@sha256:b6ef81dc8d6b6b1714d0945402ba9e8ab508c6793096f21a21fb99653ac00689'
```

The same image build by the CI:

```sh
$ curl --silent https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_csi-driver-shared-resource/46/pull-ci-openshift-csi-driver-shared-resource-master-e2e-aws-csi-driver/1430137437895528448/artifacts/build-logs/csi-driver-shared-resource.log |grep 'b6ef81dc8d6b6b1714d0945402ba9e8ab508c6793096f21a21fb99653ac00689'

2021-08-24T12:01:42.821921844Z Successfully pushed image-registry.openshift-image-registry.svc:5000/ci-op-tv0f2inb/pipeline@sha256:b6ef81dc8d6b6b1714d0945402ba9e8ab508c6793096f21a21fb99653ac00689
```

Finally, [inspecting the `DaemonSets`][gatherDaemonSets], I can confirm both the desired container image-tag and the correct arguments are informed to the container, as per:

```sh
$ curl --silent https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_csi-driver-shared-resource/46/pull-ci-openshift-csi-driver-shared-resource-master-e2e-aws-csi-driver/1430137437895528448/artifacts/e2e-aws-csi-driver/gather-extra/artifacts/daemonsets.json |jq '.items[1].spec.template.spec.containers[1] | .args, .image'
[
  "--drivername=csi.shared-resources.openshift.io",
  "--v=4",
  "--endpoint=$(CSI_ENDPOINT)",
  "--nodeid=$(KUBE_NODE_NAME)",
  "--ignorenamespace=openshift-machine-api",
  [snip]
  "--ignorenamespace=openshift-service-ca-operator"
]
"registry.build01.ci.openshift.org/ci-op-tv0f2inb/pipeline@sha256:b6ef81dc8d6b6b1714d0945402ba9e8ab508c6793096f21a21fb99653ac00689"
```

[deployScriptOutput]: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_csi-driver-shared-resource/46/pull-ci-openshift-csi-driver-shared-resource-master-e2e-aws-csi-driver/1430137437895528448/artifacts/e2e-aws-csi-driver/test/build-log.txt
[imageBuildOutput]: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_csi-driver-shared-resource/46/pull-ci-openshift-csi-driver-shared-resource-master-e2e-aws-csi-driver/1430137437895528448/artifacts/build-logs/csi-driver-shared-resource.log
[gatherDaemonSets]: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_csi-driver-shared-resource/46/pull-ci-openshift-csi-driver-shared-resource-master-e2e-aws-csi-driver/1430137437895528448/artifacts/e2e-aws-csi-driver/gather-extra/artifacts/daemonsets.json

cmd/main.go
otaviof: > I wonder if in addition to the command line arg, we document and employ a specific config map in the openshift-cluster-csi-drivers namespace that our driver uses for analogous config for his stuff, we associated keys in the config map for each of these params

I would be a good addition to this project, I think we should pursue this as a next step. I have a few consideration on what it entails and how to approach using a `ConfigMap` for this role, should we discuss during office hours?

pkg/controller/controller.go
otaviof: I would like to suggest we tackle this change in a standalone pull-request, and also discuss the details in an grooming session.

cmd/main.go
otaviof: Maybe the term "on demand" can be illustrative enough to portray the operator will populate the cache "on demand", just before mounting the actual volume.

Thus, I would suggest `--on-demand` for the flag name.

gabemontero:
> > @otaviof - as part of confirming that your `kustomize` related changes are doing what we want, please examine the `ci/prow/images` job log to locate the imagestreamtag image reference for the image built from this PR's code changes, and then go to one of the e2e job run logs on prow, and show where the deploy scripts and test output confirms that the PR test image is being used for the driver's `DaemonSet`.
> > I did a similar exercise back in [#46 (comment)](https://github.com/openshift/csi-driver-shared-resource/pull/46#issuecomment-901356817)
> 
> Okay, given the changes in `deploy.sh` and the introduction of `kustomize` is only fair to demostrante the new scripting is performing as expected.
> 
> As for the image replacement we use to tackle with `sed`, those are now working as expected by inspecting the `deploy.sh` [script output](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_csi-driver-shared-resource/46/pull-ci-openshift-csi-driver-shared-resource-master-e2e-aws-csi-driver/1430137437895528448/artifacts/e2e-aws-csi-driver/test/build-log.txt), i.e.:
> 
> ```shell
> $ curl --silent https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_csi-driver-shared-resource/46/pull-ci-openshift-csi-driver-shared-resource-master-e2e-aws-csi-driver/1430137437895528448/artifacts/e2e-aws-csi-driver/test/build-log.txt |grep -i 'patching.*image'
> 
> # Patching node-registrar image to use 'registry.build01.ci.openshift.org/ci-op-tv0f2inb/stable@sha256:b8e2d637f14e04658595393df59d99ac12284402c64c18c0a0aeecb29c70b1fb'
> # Patching node-csi-driver image to use 'registry.build01.ci.openshift.org/ci-op-tv0f2inb/pipeline@sha256:b6ef81dc8d6b6b1714d0945402ba9e8ab508c6793096f21a21fb99653ac00689'
> ```
> 
> The same image build by the CI:
> 
> ```shell
> $ curl --silent https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_csi-driver-shared-resource/46/pull-ci-openshift-csi-driver-shared-resource-master-e2e-aws-csi-driver/1430137437895528448/artifacts/build-logs/csi-driver-shared-resource.log |grep 'b6ef81dc8d6b6b1714d0945402ba9e8ab508c6793096f21a21fb99653ac00689'
> 
> 2021-08-24T12:01:42.821921844Z Successfully pushed image-registry.openshift-image-registry.svc:5000/ci-op-tv0f2inb/pipeline@sha256:b6ef81dc8d6b6b1714d0945402ba9e8ab508c6793096f21a21fb99653ac00689
> ```
> 
> Finally, [inspecting the `DaemonSets`](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_csi-driver-shared-resource/46/pull-ci-openshift-csi-driver-shared-resource-master-e2e-aws-csi-driver/1430137437895528448/artifacts/e2e-aws-csi-driver/gather-extra/artifacts/daemonsets.json), I can confirm both the desired container image-tag and the correct arguments are informed to the container, as per:
> 
> ```shell
> $ curl --silent https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_csi-driver-shared-resource/46/pull-ci-openshift-csi-driver-shared-resource-master-e2e-aws-csi-driver/1430137437895528448/artifacts/e2e-aws-csi-driver/gather-extra/artifacts/daemonsets.json |jq '.items[1].spec.template.spec.containers[1] | .args, .image'
> [
>   "--drivername=csi.shared-resources.openshift.io",
>   "--v=4",
>   "--endpoint=$(CSI_ENDPOINT)",
>   "--nodeid=$(KUBE_NODE_NAME)",
>   "--ignorenamespace=openshift-machine-api",
> /*** snip ***/
>   "--ignorenamespace=openshift-service-ca-operator"
> ]
> "registry.build01.ci.openshift.org/ci-op-tv0f2inb/pipeline@sha256:b6ef81dc8d6b6b1714d0945402ba9e8ab508c6793096f21a21fb99653ac00689"
> ```
> 
> Please consider.

Awesome thanks @otaviof looks good

/approve

for now,

between @coreydaley and I we can make the final / more detailed pass through the code and lgtm or post more comments, but seems like we are good to go

deploy/deploy.sh
```diff
@@ -16,54 +16,122 @@
...
-
-        echo >&2 "ERROR: hostpath deployment not ready after over 5min"
-        exit 1
-    fi
-    echo "$(date +%H:%M:%S) waiting for hostpath deployment to complete, attempt #$cnt"
-    cnt=$((cnt + 1))
-    sleep 10
-done
+echo "# Deploying csi driver components"
+run oc apply --kustomize ${DEPLOY_DIR}/
```
gabemontero, line 133: ok @otaviof is able to use `kustomize` via `oc apply` @adambkaplan 

so I think this closes the loop on this discussion thread

we'll use this vs. `sed`

cmd/main.go
gabemontero: @adambkaplan are you good with deferring the use of ConfigMap to configure the CSI driver to a work item outside of @otaviof 's current story here?

cmd/main.go
gabemontero: hmm ... yeah I think `--on-demand` is too far off topic

I'm still leaning toward some sort of longer name that mentions not watching because we will not update the volume after initial creation

maybe `--volume-write-create-only`

or if our default is going to be this new mode, where we are read only and you have to opt into updating volumes after their initial creation, we go with 

`--update-volume-after-create`

@adambkaplan - any opinions ?

docs/content-update-details.md
```diff
@@ -30,7 +30,10 @@ The current list of namespaces excluded from the controller's watches:
 - openshift-sdn
 - openshift-service-ca-operator
 
-The list is not yet configurable, but most likely will become so as the project's lifecycle progresses.
+The list is configurable by informing `--ignorenamespace` to the `hostpath` plugin instance. The
+plugin can also be configured with `--readonly` flag, which makes the controller read the shared
```
gabemontero, line 34: need to update this flag / text with whatever we decide here

gabemontero (commented):
next round of comments below @otaviof ... pulled in @adambkaplan into some of the discussions

my only request is that you run `shellcheck` against your new deploy.sh and if need be make the changes so it runs clean, or post evidence of the clean run

I believe there is a version of this that runs on Macs

cmd/main.go
adambkaplan: How about `--static-volume`?

The inverse feels easier to make a flag for, like `--reload-resources` to imply that we can reload the backing resources.

cmd/main.go
adambkaplan: Yes - let's split configuration via ConfigMap/file mount a separate work item.

adambkaplan: New issue in JIRA: https://issues.redhat.com/browse/BUILD-340

deploy/deploy.sh
```diff
@@ -16,54 +16,122 @@
 set -e
 set -o pipefail
 
+# when not empty it will run CSI driver on read-only mode
+READ_ONLY_MODE="${1}"
```
adambkaplan, line 20: Will probably want to update this based on what we call the flag.

cmd/main.go
otaviof: I like the idea of using `--reload-resources` and inverting the logic currently applied 👍🏼  What do you think, @gabemontero, @coreydaley?

gabemontero: I could go with `--reload-resource`.  @adambkaplan is good at names :-) 

That said, using his pivot as a building block, what does everyone think about `--refresh-resources` ?  First blush I like that even better.  But if nobody else does, I'm fine with reload.

And of course inverting the flow, so that the default is *NOT* to refresh/reload, as we've discussed in various arch/planning/etc calls is the path we want to hop on.

cmd/main.go
gabemontero: perfect thanks @adambkaplan 

feel free to resolve this or consider it processed in however fashion you prefer @otaviof - thanks

pkg/controller/controller.go
gabemontero: Agreed.  I've added a bullet to the acceptance criteria of the https://issues.redhat.com/browse/BUILD-340 item that @adambkaplan opened for us to capture this

feel free to resolve / consider this done as you see fit @otaviof - thanks

docs/content-update-details.md
gabemontero: yep either refresh/reload resources ^^

deploy/deploy.sh
gabemontero: +1

otaviof:
/retitle BUILD-264: Refresh-Resources Mode and Configurable Namespaces

docs/content-update-details.md
otaviof: Please consider latest changes.

deploy/deploy.sh
otaviof: Please consider latest changes.

Makefile
```diff
@@ -25,21 +25,36 @@ test: ## Run unit tests. Example: make test
 	go test -race -count 1 ./cmd/... ./pkg/...
 .PHONY: test
 
+# allows overwrite of node registrar image
+NODE_REGISTRAR_IMAGE ?=
+# allows overwite of CSI driver image
+DRIVER_IMAGE ?=
+# non-empty deploy mode will enable read-only mode deployment
+DEPLOY_MODE ?=
+
```
coreydaley, line 34: 1.) Should these have a default specified?
2.) These should be at the top of the file so that anyone using it will know right away what options they have

cmd/main.go
```diff
@@ -68,6 +84,17 @@ func init() {
 	rootCmd.Flags().Int64Var(&maxVolumesPerNode, "maxvolumespernode", 0, "limit of volumes per node")
 	rootCmd.Flags().StringVar(&shareRelistInterval, "share-relist-interval", "",
 		"the time between controller relist on the share resource expressed with golang time.Duration syntax(default=10m")
+	rootCmd.Flags().BoolVar(&refreshResources, "refreshresources", true, "when disabled, does watch for resource updates, only read the backing-resource just before volume mount")
```
coreydaley, line 87:
```suggestion
	rootCmd.Flags().BoolVar(&refreshResources, "refreshresources", true, " watch for resource updates")
```

cmd/main.go
```diff
@@ -68,6 +84,17 @@ func init() {
...
+	rootCmd.Flags().StringSliceVar(&ignoredNamespaces, "ignorenamespace", []string{}, "namespaces ignored by the controller")
```
coreydaley, line 88:
```suggestion
	rootCmd.Flags().StringSliceVar(&ignoredNamespaces, "ignorenamespace", []string{}, "Specify a namespace to be ignored by the controller")
```

pkg/controller/controller.go
```diff
@@ -63,37 +68,11 @@ func NewController(shareRelist time.Duration) (*Controller, error) {
...
-			"openshift-cluster-storage-operator",
-			"openshift-cluster-version",
-			"openshift-image-registry",
-			"openshift-machine-config-operator",
-			"openshift-sdn",
-			"openshift-service-ca-operator",
-		}
-		for _, ns := range namespaces {
+		for _, ns := range ignoredNamespaces {
+			klog.V(4).Infof("namespace '%s' is ignored", ns)
```
coreydaley, line 75:
```suggestion
			klog.V(4).Infof("namespace '%s' is being ignored", ns)
```

Makefile
otaviof: Those are variables previously used on `deploy.sh`, now also exposed by `Makefile` as well.

otaviof: The `DEPLOY_MODE` is new, and I'm updating the comment to make clear what's its objective. Let me know if it's clear.

cmd/main.go
gabemontero: I'm good with where we are here now ... thanks everyone

docs/content-update-details.md
```diff
@@ -30,7 +30,12 @@ The current list of namespaces excluded from the controller's watches:
 - openshift-sdn
 - openshift-service-ca-operator
 
-The list is not yet configurable, but most likely will become so as the project's lifecycle progresses.
+The list is configurable by informing `--ignorenamespace` to the `hostpath` plugin instance. The
+plugin can also be configured with `--refreshresources` flag, which makes the controller keep a warm
+cache of `ConfigMap` and `Secret` objects, and as they change the controller will follow the updates.
+When `--refreshresources` is disabled (i.e. `--refreshresources=false`), the controller will read
+the backing-resource (`.spec.backingResource`) just before mounting the volume instead of keeping a
+warm cache.
```
gabemontero, line 38: This looks good with the current implementation, but it reminded me of something.  Something not explicitly called out in the Jira as I  revisited it, but something I always had in mind when we first created this story.

Namely, if refreshresources is disabled, we force the volume to be read only.
1) the code check would be around https://github.com/openshift/csi-driver-shared-resource/blob/7c62db45752262172549a6f4ee5fee3c605606f2/pkg/hostpath/nodeserver.go#L163-L165
2) I don't see why for the scenarios we are considering we would allow the consuming pod to change the contents we mounted, or add new files to our mount
3) It defers the need of the atomic writer stuff from k8s/k8s as we would not have to ever coordinate with the pod 
4) remember, we claim to not be a "generic" csi volume provider, so ignoring user options like read only is a legitimate stance

@adambkaplan WDYT ?

Makefile
gabemontero: the comment update looks good @otaviof 

but I agree with @coreydaley that it should be moved toward the top, above `all: ....`

same with the TEST_SUITE and TEST_TIMEOUT that I think @adambkaplan recently added

thanks

gabemontero (changes requested):
One minor ask, one suggested change that admittedly needs to be discussed, where I could see us going either way @otaviof and we need to discuss either here on or video conf with @adambkaplan 

Also, I see the e2e commit for when refresh resources is disabled, and I see the unit test changes, but I was expecting e2e test changes in the go files where minimally the e2e tests that changed the share when refresh resources was disabled were not executed when we did `test-e2e-no-refreshresources` per your Makefile update, as well as making sure content was not updated

i.e. a new `TEST_SUITE`

Were you planning those in a subsequent PR?

gabemontero:
not sure why it removed approved  .... some git intergration we don't have elsewhere 

/approve

deploy/csi-hostpath-plugin.yaml
```diff
@@ -70,6 +70,30 @@ spec:
...
+            - --ignorenamespace=openshift-config-operator
+            - --ignorenamespace=openshift-etcd-operator
+            - --ignorenamespace=openshift-apiserver-operator
+            - --ignorenamespace=openshift-cluster-csi-drivers
+            - --ignorenamespace=openshift-cluster-storage-operator
+            - --ignorenamespace=openshift-cluster-version
+            - --ignorenamespace=openshift-image-registry
+            - --ignorenamespace=openshift-machine-config-operator
+            - --ignorenamespace=openshift-sdn
+            - --ignorenamespace=openshift-service-ca-operator
```
coreydaley, line 96: Is this list going to have to get updated every time a new openshift-* namespace comes along? Couldn't we exclude all openshift-* namespaces with a regexp in the code somewhere if that is what needs to be done?

pkg/controller/controller.go
```diff
@@ -63,37 +68,11 @@ func NewController(shareRelist time.Duration) (*Controller, error) {
...
-			"openshift-machine-config-operator",
-			"openshift-sdn",
-			"openshift-service-ca-operator",
-		}
-		for _, ns := range namespaces {
+		for _, ns := range ignoredNamespaces {
+			klog.V(4).Infof("namespace '%s' is being ignored", ns)
 			nsfs := fmt.Sprintf(namespaceFieldSelector, ns)
 			if len(fsString) == 0 {
 				fsString = nsfs
```
coreydaley, line 78:
```suggestion
		for _, ns := range ignoredNamespaces {
			klog.V(4).Infof("namespace '%s' is being ignored", ns)
			options.FiledSelector = fmt.Sprintf(namespaceFieldSelector, ns)
		}
```
Is there something I am missing, or can this all be refactored to the above suggestion?

gabemontero:
For the rotating cert scenario we may need to watch certain openshift
namespaces

I at least am fine with deferring more sophisticated regex / other means of
expression in a separate PR

pkg/controller/controller.go
```diff
@@ -63,37 +68,11 @@ func NewController(shareRelist time.Duration) (*Controller, error) {
 	// informers.NewSharedInformerFactoryWithOptions, but we restrict OpenShift
 	// "system" namespaces with chatty configmaps like the leaderelection related ones
 	// that are updated every few seconds
-	//TODO make this list externally configurable
```
coreydaley, line 66:
```
tweakListOptions := internalinterfaces.TweakListOptionsFunc(func(options *metav1.ListOptions) {
        ignored := []string{}
	for _, ns := range ignoredNamespaces {
		ignored = append(ignored, fmt.Sprintf("metadata.namespace!=%s", ns))
	}
        options.FieldSelector = strings.Join(ignored, ",")
})
```

coreydaley: This would be cleaner than all the appending strings stuff.

deploy/csi-hostpath-plugin.yaml
gabemontero: For the rotating cert scenario we may need to watch certain openshift
namespaces

I at least am fine with deferring more sophisticated regex / other means of
expression in a separate PR

pkg/controller/controller.go
gabemontero: true ... technically a tech debt item given @otaviof did not change that existing bit

I'm fine if he want to take that on now, or somebody does this as a follow up PR

otaviof: Thanks for the heads up, @coreydaley, it's more readable this way. However, I would argue this change should be tackled in a subsequent PR, would be "a good first issue" by the way.

deploy/csi-hostpath-plugin.yaml
otaviof: I also think we should take this in a standalone pull-request, I would like discuss this requirement further.

otaviof:
> Also, I see the e2e commit for when refresh resources is disabled, and I see the unit test changes, but I was expecting e2e test changes in the go files where minimally the e2e tests that changed the share when refresh resources was disabled were not executed when we did `test-e2e-no-refreshresources` per your Makefile update, as well as making sure content was not updated

@gabemontero, in my view the `--refreshresources=false` is orthogonal to the `TEST_SUITE`. So, as a follow up step I was about to update the CI configuration to also run the `normal` test-suite with this flag enabled. Would you second this move?

I've tested this mode against all test suites successfuly. And on this PR we have unit-testing to assert the HostPathDriver works as expected when a client is informed: https://github.com/otaviof/csi-driver-shared-resource/blob/1406188d39046dc4de5a92c131682cadefcc7390/pkg/hostpath/hostpath_test.go#L544-L645

docs/content-update-details.md
otaviof: I think it's important to highlight that when `--refreshresources` is disabled, the data on the object-cache is refreshed just before the volume is mounted. Therefore, every time the volume is mounted the cache-key would get in sync with the apiserver again.

gabemontero: we have both "short living" pods like builds and will have "long running" pods which will be comsuming config maps with certs

in those long running cases, the volumes would only be mounted once

but while a good distinction to note @otaviof I think it is still orthogonal to the classic atomic write k8s secret stuff 

there are 2 main motivators IMO for this feature's original creation
1) scaling benefits from not employing a watch
2) defaulting to read only to avoid atomic write hoopla when we know we won't update content after pod creation

gabemontero:
> > Also, I see the e2e commit for when refresh resources is disabled, and I see the unit test changes, but I was expecting e2e test changes in the go files where minimally the e2e tests that changed the share when refresh resources was disabled were not executed when we did `test-e2e-no-refreshresources` per your Makefile update, as well as making sure content was not updated
> 
> @gabemontero, in my view the `--refreshresources=false` is orthogonal to the `TEST_SUITE`. So, as a follow up step I was about to update the CI configuration to also run the `normal` test-suite with this flag enabled. Would you second this move?

I'm not sure I agree it is orthogonal @otaviof , but I'll defer "officially" declaring that objection until I see more precisely what your openshift/release change is

If I see something equivalent to both
- running normal with  `--refreshresources=false` 
- running normal with  `--refreshresources=true`

I'll be satisfied.  I did refresh my memory, and normal only deals with deletion of shares, so I agree its tests as is should work regardless of the refreshresources setting. 
> 
> I've tested this mode against all test suites successfuly. And on this PR we have unit-testing to assert the HostPathDriver works as expected when a client is informed: https://github.com/otaviof/csi-driver-shared-resource/blob/1406188d39046dc4de5a92c131682cadefcc7390/pkg/hostpath/hostpath_test.go#L544-L645

Yes I realize the tests are currently passing.

docs/content-update-details.md
otaviof: @gabemontero please consider the latest changes. Now on every volume mount when refresh-resources is not enabled, it will be a read-only volume.

gabemontero (approved):
/lgtm

pkg/controller/controller.go
otaviof: @coreydaley I'm bringing that up on #60, thanks for the suggestion!

PR #50, opened 2021-09-13
BUILD-234: Add cross platform build support in Makefile

adambkaplan:
- Allow target `GOOS` and `GOARCH` to be configurable with
  `TARGET_GOOS` and `TARGET_GOARCH` variables
- Add race detection only if compiling for amd64.
- Remove ldflags that break compilation on s390x.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,17 @@ TEST_SUITE ?= normal
 # ent-to-end test timeout
 TEST_TIMEOUT ?= 30m
 
-LDFLAGS ?= '-extldflags "-static"'
+TARGET_GOOS ?= $(shell go env GOOS)
+TARGET_GOARCH ?= $(shell go env GOARCH)
+
+# For golang 1.16, race detection is only widely supported for amd64 architectures (linux, windows, darwin, and freebsd).
+# Race detection for ARM is only currently supported for linux (no darwin or windows support yet).
+# s390x and ppc64le do not support race detection at present.
+ifeq ($(TARGET_GOARCH), amd64)
+	RACE = -race
+endif
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -72,7 +72,7 @@ verify: ## Run verifications. Example: make verify
 .PHONY: verify
 
 build: ## Build the executable. Example: make build
-	go build -a -mod=vendor -race -ldflags $(LDFLAGS) -o _output/csi-driver-shared-resource ./cmd
```
gabemontero, line 75: could you add a `build-race` make target that still let's one do the race tests when they are on platforms that support it @adambkaplan

otaviof: I just sent a [small PR to tackle `-race` flag](https://github.com/adambkaplan/openshift-csi-driver-shared-resource/pull/1) on different platform, so we can keep a single `build` target on the `Makefile`. Please consider.

gabemontero: oooh even better :-) ...... thanks @otaviof 

and got to admit, I employed such platform checks quite a bit in my now distant past, programming in C and C++, and totally forgot about them; in my defense, @otaviof is much younger than me and undoubtedly has a better memory :-)

otaviof: Thanks, Gabe! Granted some "creative stealing" to find a good example out there. :-)

adambkaplan:
@gabemontero @otaviof updated Makefile such that we can compile on Power and Z.

I may want to follow up with a `verify` test which cross-compiles the CSI driver on amd64, arm, ppc64le, and s390x

Makefile
adambkaplan: done!

gabemontero:
unrelated flake

/retest

/lgtm

PR #54, opened 2021-09-25
Build 238: share metrics exposure, new monitoring objects

alicerum:
This commit adds metrics exposure (`/metrics` endpoint and shares metrics collector) for the `openshift_csi_shares_total` metrics.
Also included are new deployment objects for the `prometheus` operator, so that this new metric is scraped correctly by OpenShift.


pkg/metrics/metrics.go
```diff
@@ -0,0 +1,126 @@
...
+	sharesCountQuery = sharesSubsystem + separator + "total"
+
+	MetricsPort = 6000
+)
+
+var (
+	sharesCountDesc = prometheus.NewDesc(
+		sharesCountQuery,
+		"Counts shares by kind of resource and namespace",
+		[]string{"source_namespace", "kind"},
```
gabemontero, line 26: with the recent breaking out of the shared crd to sharedConfigMap and sharedSecret crd, we should have different metrics for eaach of those new types now, and hence you lose the "kind" label @alicerum 

I would also expect a "source_name" label

Of course adjust the descriptions accordingly

I think it is also important to explain what source ns/name mean .... otherwise it is natural for consumers to get confused by their typical understanding of what those labels mean with other metrics, as @jan--f  did

gabemontero (changes requested):
do the separate metrics for shared secret and shared config map @alicefr and then I'll come back and do a more detailed review

also, I am expecting unit tests for metrics.go minimally .... server.go is possible but optional

see examples with our other components like 

https://github.com/openshift/openshift-controller-manager/tree/master/pkg/build/metrics/prometheus

https://github.com/openshift/cluster-samples-operator/tree/master/pkg/metrics

https://github.com/openshift/cluster-image-registry-operator/tree/master/pkg/metrics

pkg/metrics/metrics.go
alicerum: @gabemontero i have number of mounts prepared as a different PR, wanted to push them separately. I think I had this discussion with @adambkaplan , that I should create PR as soon as I had one metrics implemented.
I'll separate the metrics for config maps and secrets. We had a discussion with @adambkaplan today, that I might want to move all this to https://github.com/openshift/csi-driver-shared-resource-operator due to a number of reasons, including controller running in a daemon set (on every node). What do you think?

jan--f:
Makes sense to me.

I did consider making a change to the `kind` label. The naming is not ideal and, since it has a finite value set, splitting it up into two metrics is an option. Whether this should be a label or two metrics again depends on the intended use case. One metric with a label is nicer to aggregate (`sum(openshift_csi_shares_total)` instead of `openshift_csi_secret_shares_total + openshift_csi_configmap_shares_total`). If that is an unlikely usage of the metrics, two metrics are better then a label.

:+1: on the namespace related label.

gabemontero:
new unit tests look good @alicerum  ... and yeah, as I previously noted, just having unit tests for metrics.go and not server.go is OK at least for now, based on my prior experience with our other operators and the usefulness / viability of that test

but we'll probably want to solicit input from @adambkaplan on that as well

otherwise, am I correct that success/failure metrics is still in your queue, we just need to talk about it in office hours?

thanks

go.mod
```diff
@@ -3,9 +3,10 @@ module github.com/openshift/csi-driver-shared-resource
 go 1.14
 
 require (
+	github.com/blang/semver v3.5.1+incompatible
 	github.com/container-storage-interface/spec v1.3.0
 	github.com/kubernetes-csi/csi-lib-utils v0.7.0
-	github.com/openshift/api v0.0.0-20210702094102-edc45a4714d6
+	github.com/prometheus/client_golang v1.11.0
```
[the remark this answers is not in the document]
alicerum: I think it got removed during `go mod tidy` or similar for whatever reason. I'll do a little research about it.

PR #57, opened 2021-10-01
[WIP] BUILD-284 Use openshift/api and openshift/client-go for clientset, informers, and listers

coreydaley:


gabemontero:
realizing this is still WIP and the openshift/api PR has not been fully approved/merge, so changes with the API are not out of the question, but the relative simplicity of these changes so far make me smile :-)

hack/generate-release-yaml.sh
```diff
@@ -20,6 +22,6 @@ for FILE in ${FILES[@]}
 do
 
 	echo -e "\n---\n" >> release.yaml
-	cat deploy/${FILE} >> release.yaml
+	cat ${FILE} >> release.yaml
```
gabemontero, line 25: yeah this most likely will go away once we intergrate with the CSO, but better to keep it sensible in the interim

pkg/cache/common.go
```diff
@@ -19,8 +17,8 @@ func GetKey(o interface{}) string {
 	return obj.GetNamespace() + ":" + obj.GetName()
 }
 
-func BuildKey(r sharev1alpha1.ResourceReference) string {
-	return r.Namespace + ":" + r.Name
+func BuildKey(namespace, name string) string {
+	return namespace + ":" + name
```
gabemontero, line 21: feels like a "non-required" change but OK :-)

pkg/client/client.go
```diff
@@ -94,17 +95,17 @@ func initClient() error {
 	return nil
 }
 
-func ExecuteSAR(shareName, podNamespace, podName, podSA string, kind sharev1alpha1.ResourceReferenceType) (bool, error) {
+func ExecuteSAR(shareName, podNamespace, podName, podSA string, kind consts.ResourceReferenceType) (bool, error) {
```
gabemontero, line 98: right this came up in the openshift/api review

PR #60, opened 2021-10-11
[WIP] BUILD-340: Configuration over ConfigMap

otaviof:


docs/config.md
```diff
@@ -0,0 +1,42 @@
+# Configuration
+
+The CSI driver instance is configured by a combination of command-line flags and a YAML configuration
+file. You can use the following flags:
+
+- `--config`: path to the YAML configuration file, optional;
+- `--nodeid`: the current Kubernetes node identifier, must be the same than the node it's running on
+- `-—refreshresources`: toggles the operation mode, when enabled it continuously watch for `Secret`
```
gabemontero, line 8: OK I'm struggling a bit with the presentation here, and the distinction of what is available in general from the command line, vs. what we envision a cluster admin wanting to manipulate.

I'm going to briefly present things in my words @otaviof and then let's see how we can reconcile that with what you are trying to convey here.

1) I envision initially surfacing via the config map the only the subset of command line options we think the cluster admin will want to manipulate / we think the *should* manipulate

I see 3 of those:
- share-rellist-interval
- refreshresources
- ignorenamespace

From a code reference perspective

https://github.com/openshift/csi-driver-shared-resource/blob/c8dab4fb90277123868abe8f05086fa48ed1f2d5/cmd/main.go#L86-L89

2) yes, we currently have the other 4 command line options at https://github.com/openshift/csi-driver-shared-resource/blob/c8dab4fb90277123868abe8f05086fa48ed1f2d5/cmd/main.go#L81-L85

- endpoint
- drivername
- nodeid
- maxvolumespernode

to be fair, it is not readily apparent to anyone but me, since I am the one who "originated" this driver, but these and config options that existing in the upstream, generic CSI hostpath/secret share volume drivers I originally based csi-driver-share-resource from.

most likely, as we move toward tech preview and GA, we'll either get rid of these entirely, or we'll keep them hidden, in reserve, in case user interaction and feedback informs us of some tuning for those we are currently not anticipating.

So, digest all this ^^ @otaviof, let's talk "voice to voice" if need be, and then let's see how we change this up

also to be fair, none of this nuance is captured in https://issues.redhat.com/browse/BUILD-340

pkg/config/manager.go
```diff
@@ -0,0 +1,66 @@
...
+		klog.Errorf("Reading configuration-file '%s': '%#v'", m.cfgFilePath, err)
+		return false
+	}
+	sum := md5.Sum(payload)
+	return m.md5sum != hex.EncodeToString(sum[:])
+}
+
+// LoadConfig read the local configuration file, make sure the current contents are summed, so we can
+// assert if there are changes later on.
+func (m *Manager) LoadConfig() (*Config, error) {
```
gabemontero, line 40: OK seeing all this, I think this lends itself to my option 1) in the comment I made in main.go @otaviof 

WDYT?

gabemontero (changes requested):
OK @otaviof after seeing this I think I better understand some of the points you were making in office hours / scrum.

And that said, taking in some of the additional detail I noted below that would have been ideal to go over during planning, I think this PR is on the right path to being the changes needed in this repo that could go along with the changes needed in https://github.com/openshift/csi-driver-shared-resource-operator where that watches / reacts to changes in the configmap, initiates restarts on this driver's daemonset / pods, follow some of the OCM-O / OCM "patterns" we can learn from etc.

So I'll wait to hear how my comments below land with you, if some voice to voice conversations make sense, how other members of the team react to my input, and we'll go from there.

Thanks!

docs/config.md
otaviof: Thanks for the explanation ([here](https://github.com/openshift/csi-driver-shared-resource/pull/60#discussion_r729050639)) about the historical origins of those flags, it gives the input need so we can define a clear distinction of what goes where.

I'm following your suggestion, please consider latest changes.

cmd/main.go
```diff
@@ -79,14 +92,9 @@ func init() {
...
 
 	rootCmd.Flags().AddGoFlagSet(flag.CommandLine)
-	rootCmd.Flags().StringVar(&endPoint, "endpoint", "unix://tmp/csi.sock", "CSI endpoint")
-	rootCmd.Flags().StringVar(&driverName, "drivername", string(operatorv1.SharedResourcesCSIDriver), "name of the driver")
-	rootCmd.Flags().StringVar(&nodeID, "nodeid", "", "node id")
-	rootCmd.Flags().Int64Var(&maxVolumesPerNode, "maxvolumespernode", 0, "limit of volumes per node")
-	rootCmd.Flags().StringVar(&shareRelistInterval, "share-relist-interval", "",
-		"the time between controller relist on the share resource expressed with golang time.Duration syntax(default=10m")
+	rootCmd.Flags().StringVar(&cfgFilePath, "config", "/var/run/configmaps/config/config.yaml", "configuration file path")
+	rootCmd.Flags().StringVar(&nodeID, "nodeid", "", "Kubernetes node identification")
```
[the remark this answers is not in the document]
otaviof: My [previous comment](https://github.com/openshift/csi-driver-shared-resource/pull/60#discussion_r730694753) addresses the configuration attributes, what stays as flags and what goes to the configuration file, please consider.

> We either
> 
>     1. always just read from the config map
> 
>     2. always depend on the csi-driver-shared-resource-operator to map into changes to our DaemonSet and the command line options specfied there ... i.e. https://github.com/openshift/csi-driver-shared-resource/blob/c8dab4fb90277123868abe8f05086fa48ed1f2d5/deploy/csi-hostpath-plugin.yaml#L68-L96
> 
> 
> My initial thought is 1) is better, but not having done the end to end coding on this you have, I'll defer to you if you have insight based on your prototyping that says otherwise.

I agree option `1` is better, it's also the subject of latest changes as well. The configuration file is always mounted from the `ConfigMap` rolled out with the driver, when it changes, the driver is restarted too.

cmd/main.go
```diff
@@ -45,7 +44,14 @@ var rootCmd = &cobra.Command{
 		var kubeClient kubernetes.Interface
 		var err error
 
-		if !refreshResources {
+		cfgManager := config.NewManager(cfgFilePath)
+		cfg, err := cfgManager.LoadConfig()
+		if err != nil {
+			fmt.Printf("Failed to load configuration file '%s': %s", cfgFilePath, err.Error())
+			os.Exit(1)
```
gabemontero, line 51: OK this exit caught my eye.  I was assuming we could go with defaults for our three config fields and continue to function, with appropriate logging, if the mounted config file for the config map was missing.

So I went back to the Jira and looked at the OCM-O pointer, and tried to track down what it does.

So starting with https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/pkg/operator/sync_openshiftcontrollermanager_v311_00.go#L183-L189

Which took me to https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/bindata/v3.11.0/openshift-controller-manager/ds.yaml#L44-L45

Skip a few steps to 

https://github.com/openshift/openshift-controller-manager/blob/fd48f0a784383782b1ac58d95e732f8df516b844/pkg/cmd/openshift-controller-manager/cmd.go#L65-L68

I do see this:  https://github.com/openshift/openshift-controller-manager/blob/fd48f0a784383782b1ac58d95e732f8df516b844/pkg/cmd/openshift-controller-manager/cmd.go#L95-L98

So this is following the same "precedence" that was established in OCM-O etc.

I still wonder if we can go over and above, and assume defaults here too. 

What do you think @otaviof ?  Among other things, you doc file as I currently read it that we will assume defaults if the file is missing.

Lastly, also please add a TODO comment that we need to make sure how we report the lack of the config map here lines up with how our operator manages us with the.  Presumably the opeartor will also note the lack of configmap and we can decide there how it treats that with respect to reporting status to the CSI.  Some for of "degraded".  Or do we create a metric / alert for this condition perhaps.

cmd/main.go
```diff
@@ -122,6 +126,18 @@ func runOperator() {
 	}
 }
 
+// watchForConfigChanges keeps checking if the informed configuration has changed, and in this case
+// makes the operator exit. The new configuration should take place upon new instance started.
+func watchForConfigChanges(mgr *config.Manager) {
+	for {
+		if mgr.ConfigHasChanged() {
```
gabemontero, line 133: this is fine for now for prototype testing, but add a TODO comment that when the operator changes are in place, we quit watching and reacting to updates in this driver, and will defer to the operator to watch for config changes, and restart our entire daemonset when it sees them, as I noted in my comment https://github.com/openshift/csi-driver-shared-resource/pull/60#pullrequestreview-779831515 where we follow / employ the library-go related utilities like OCM-O / OCM

If there was something already analogous to this in OCM, I'm missing it.  Do you have such a pointer?

docs/config.md
```diff
@@ -0,0 +1,37 @@
...
+shareRelistInterval: 10m
+
+# toggles actively watching for resources, when disabled it will only read objects before mount
+refreshResources: true
+
+# list of namespace names ignored
+ignoredNamespaces: []
+```
+
+When the file is not present, the driver assumes default values instead. And, when the configuration
```
gabemontero, line 32: ok here you go :-) .... you said we'll assume defaults, good.

but you've got that exit up top in main.go if you could not load the config.

And I'm not seeing how in the code it is being achieved yet ... if it is suppose to be the new yaml in the deploy dir, I'm not seeing that as the best way (vs. the code seeding things if the file is missing).

docs/config.md
```diff
@@ -0,0 +1,37 @@
...
+contents change,  it restarts after a couple second, allowing Kubernetes to restart it back again,
+with updated configs.
+
+During the rollout, a `config` named `ConfigMap` is created with default configuration values, in the
```
gabemontero, line 36: you mean `ConfigMap` named `config` I think

that said, let's get a little more precise in our name, as we'll be sharing a namespace with the other CSI drivers

go with `csi-driver-shared-resource-config`

deploy/config.yaml
```diff
@@ -0,0 +1,26 @@
+---
+ignoredNamespaces:
```
gabemontero, line 2: Is this suppose to be how we assume defaults, per your doc update?  If so, I'm not sure that is the best way to do it.

Or is this just a ease of use thing for our development flow via `make deploy` that bypasses the operator?

If so, we should have settings for the other 2 fields, for clarity as much as anything, even if defaults are gleamed in the code itself.

gabemontero (changes requested):
A lot of the core details seem fine @otaviof, I just have a couple of macro items around assuming defaults, and if the mounted config map is missing.

Also, I think the unit tests are good for now.  You may have already been thinking this, but I would expect and e2e defined here for actually verifying consumption of a config change / restart, but it makes sense to define that e2e here once the operator wiring gets worked.

cmd/main.go
otaviof: Have you considered what the LoadConfig does, and [how it may error out](https://github.com/otaviof/csi-driver-shared-resource/blob/04372abf4f3b05cf1e01395d64cf87207f8adf61/pkg/config/manager.go#L48-L59)?

docs/config.md
otaviof: You may want to consider [unit-testing](https://github.com/otaviof/csi-driver-shared-resource/blob/04372abf4f3b05cf1e01395d64cf87207f8adf61/pkg/config/manager_test.go#L26-L45) to assert there are defaults, and those defaults can be overwritten by configuration file.

deploy/config.yaml
otaviof: Yes, the `config.yaml` is in fact the configuration file embedded in the `ConfigMap`, therefore that's the place to record the default values for OpenShift related things. Other default attributes the operator needs to assume in order to function correctly [is here](https://github.com/otaviof/csi-driver-shared-resource/blob/04372abf4f3b05cf1e01395d64cf87207f8adf61/pkg/config/config.go#L33-L40).

cmd/main.go
gabemontero: Hmmm ... as I read the acceptance criteria in BUILD-340, it says "The shared resource CSI driver **operator** watches for create/update on this ConfigMap and processes config changes dynamically, or considers restart if appropriate."

Now, if you want to work with @adambkaplan to break up BUILD-340, then sure let's discuss that in our team meetings.

But as I read that acceptance criteria, your watching for config changes needs to move out of here and into the opeator.

deploy/config.yaml
gabemontero: https://github.com/otaviof/csi-driver-shared-resource/blob/04372abf4f3b05cf1e01395d64cf87207f8adf61/pkg/config/config.go#L33-L40 was the bit of code I was missing .... thanks @otaviof

docs/config.md
gabemontero: OK the unit test did not help with my "how" question, your pointer in your comment https://github.com/openshift/csi-driver-shared-resource/pull/60#discussion_r731093815 did ... so I'm good wrt this.

cmd/main.go
gabemontero: OK I think I have pieced together what you are getting at, and I appreciate the code links.

To state a bit more precisely and specificity:  
1) if the config file from the mounted config map is removed, assume defaults, no error returned
2) if that file is somehow corrupt, we don't want to assume defaults, since the user may have intended to override the defaults

Perhaps that should have been obvious to this reviewer.  A comment capturing 2) would have helped me, but I won't insist on it :-)

Now, as part of my original comment for a TODO comment around following up how the operator knows that this driver is giving up because the config file is corrupt is still warranted.

We can then sort out the resolution of that TODO as we sort out how much of the operator side you complete with BUILD-340.  Perhaps the TODO remains until any potential follow up to BUILD-340 is done.

cmd/main.go
gabemontero: OK I don't remember this precise element coming up in those prior discussions, about deferring on making a change in the operator.  I remember discussion on the mechanics on how to do it.

Other than trying again to sort out those details, and deciding what is done now vs. later, I don't have any other suggestions.

pkg/config/config_test.go
```diff
@@ -0,0 +1,26 @@
+package config
+
+import (
+	"testing"
+	"time"
+)
+
+func TestConfig_GetShareRelistInterval(t *testing.T) {
```
gabemontero, line 8: OK @otaviof wrt the need for an e2e in addition to your unit tests here, I think it is valid to have tests verify that 
1) our daemonset pod's exit successfully and k8s restarts our pods
2) say you add a new namespace to ignore; you could have pods that were provisioned with read/write volumes to a secret or configmap from that namespace, and prove they see updates to the config map or secret; you then change the config to start ignoring that namespace; the driver restarts; you then can validate that the pod no longer sees additional updates from the config map or secret from the namespace that is now ignored  .... I could envision a cluster admin wanting to do this as a means of "freezing" the content in question from consuming pods, or other scenarios we have not envisioned yet .... it is not just that the setting in the config file is verified as changed.  It is the propagation of the change end to end.  And yes, we have tests to verify those specific settings.  But one of the common reasons for e2e testing is to validate that multiple features work in conjunction.
3) the same scenario could be augmented to a global level, and we freeze updates everywhere by setting refresh resources to false

I'd be curious if @adambkaplan or @coreydaley either a) feel I asking more than what should be minimally expected for this story.  If I'm the sole opinion who things such e2e's are a good idea, I'll defer.  I'm also curious if either of them can think of other e2e type scenarios with respect to config change that I am omitting here.

gabemontero: But I don't see this existing unit test covering any of ^^ above

pkg/config/manager_test.go
```diff
@@ -0,0 +1,67 @@
+package config
+
+import (
+	"reflect"
+	"testing"
+)
+
+// TestConfig_DefaultConfig asserts the default configuration attributes is in use when a
+// non-existing configuration file is informed.
+func TestConfig_DefaultConfig(t *testing.T) {
```
gabemontero, line 10: sorry forgot to note @otaviof that the second unit test you added best as I can see address things I noted per https://github.com/openshift/csi-driver-shared-resource/pull/60#discussion_r732178310

gabemontero:
Shoot forgot one more point

Is kustomize pervasive enough / consumable enough for our support / customer level based on our experience?

That avoids the delete / recreate option ^^

deploy/deploy.sh
```diff
@@ -79,34 +82,34 @@ function kustomize_set_image () {
...
-  {
-    "op": "add",
-    "path": "/spec/template/spec/containers/1/args/-",
-    "value": "${ARG}"
-  }
-]
-EOS
+# adds the settings to generate the configuration ConfigMap, taking in consideration extra flag for
+# refresh-resources mode.
+function kustomize_config () {
```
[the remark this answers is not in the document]
gabemontero: Last update for today :-)  ... though hopefully a good one.

I think I've sorted out enough on the CI side and BUILD-256 that I can work with with deploy.sh and the Makefile as you have it here.

PR #63, opened 2021-10-14
WIP: Use hostpath volume for driver's data

akram:
This is supposed to fix the failure on restart of the CSI driver when using shared volumes backed by volumes stored on tmpfs.

### DIFF
--- a/examples/csi-app.yaml
+++ b/examples/csi-app.yaml
@@ -21,7 +21,7 @@ spec:
             touch /data/foo
             ls -laZ /data
             echo "calling sleep"
-            sleep 120
+            sleep 10
           done
   volumes:
     - name: my-csi-volume
--- a/examples/csi-app.yaml
+++ b/examples/csi-app.yaml
@@ -7,7 +7,7 @@ spec:
   serviceAccountName: default
   containers:
     - name: my-frontend
-      image: quay.io/quay/busybox
+      image: registry.access.redhat.com/ubi8/ubi
       volumeMounts:
         - mountPath: "/data"
           name: my-csi-volume
@@ -17,16 +17,16 @@ spec:
         - |
           while true
           do
-            ls -la /data
+            ls -laZ /data
             touch /data/foo
-            ls -la /data
+            ls -laZ /data
             echo "calling sleep"
             sleep 120
           done
   volumes:
     - name: my-csi-volume
       csi:
-        #readOnly: true
+        readOnly: true
         driver: csi.sharedresource.openshift.io
         volumeAttributes:
           sharedConfigMap: my-share
--- a/pkg/hostpath/constants.go
+++ b/pkg/hostpath/constants.go
@@ -17,7 +17,8 @@ const (
 	CSIEphemeral                       = "csi.storage.k8s.io/ephemeral"
 	SharedConfigMapShareKey            = "sharedConfigMap"
 	SharedSecretShareKey               = "sharedSecret"
-	//anchorDir                          = "anchor-dir"
-	bindDir                            = "bind-dir"
+	AnchorDir                          = "anchor-dir"
+	TmpFsType                          = "tmpfs"
+	BindDir                            = "bind-dir"
 	mountAccess             accessType = iota
 )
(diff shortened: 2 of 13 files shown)


examples/csi-app.yaml
```diff
@@ -7,7 +7,7 @@ spec:
   serviceAccountName: default
   containers:
     - name: my-frontend
-      image: quay.io/quay/busybox
+      image: registry.access.redhat.com/ubi8/ubi
```
gabemontero, line 10: not that this is "bad", but why did you feel the need to change this?

pkg/hostpath/constants.go
```diff
@@ -17,7 +17,7 @@ const (
 	CSIEphemeral                       = "csi.storage.k8s.io/ephemeral"
 	SharedConfigMapShareKey            = "sharedConfigMap"
 	SharedSecretShareKey               = "sharedSecret"
-	anchorDir                          = "anchor-dir"
+	//anchorDir                          = "anchor-dir"
```
gabemontero, line 20: yeah I know my earlier POC branch just commented this out, but you can feel free to just delete this

pkg/hostpath/mount.go
```diff
@@ -52,15 +52,15 @@ type ReadWriteMany struct {
 
 func (m *ReadWriteMany) makeFSMounts(anchorDir, intermediateBindMountDir, kubeletTargetDir string, mounter mount.Interface) error {
 	options := []string{}
-	if err := mounter.Mount(anchorDir, kubeletTargetDir, "tmpfs", options); err != nil {
+	if err := mounter.Mount("GGM", kubeletTargetDir, "tmpfs", options); err != nil {
```
gabemontero, line 55: yeah I mentioned this on our video conf, though quite possibly very rapidly, to the point it was easy to miss ;-)

but we of course want to change this to a free form string relevant to our volume

IMO, a combination of pod name, pod namespace, volume ID, minimally
the share type and name could also be helpful, but I'm undecided, as perhaps that is too much

deploy/csi-hostpath-plugin.yaml
```diff
@@ -56,7 +56,7 @@ spec:
               name: socket-dir
             - mountPath: /registration
               name: registration-dir
-            - mountPath: /csi-data-dir
+            - mountPath: /run/csi-data-dir
```
gabemontero, line 59: Jan's comment was specifically /run/csi driver name,

i.e. https://github.com/openshift/csi-driver-shared-resource/blob/master/vendor/github.com/openshift/api/operator/v1/types_csi_cluster_driver.go#L55

of course you cannot use that constant in yaml, but use the string that constant is set to

examples/csi-app.yaml
gabemontero: nope that is sufficient justification for me ... let's continue with ubi image .... that is "more strategic" from a RH perspective anyway

pkg/hostpath/mount.go
akram: The value format that you are mentionning @gabemontero is already something returned by `func (hp *hostPath) getVolumePath(volID string, volCtx map[string]string)` which uses `anchorDir` as a constant to set all the anchor dirs under the same directory.
I will jtry to use the returned value of this method to mount it there. That should make the job.

akram:
@gabemontero tests are failing now: could it be related to https://github.com/openshift/csi-driver-shared-resource-operator/pull/24 not being merged yet ?

gabemontero:
> @gabemontero tests are failing now: could it be related to [openshift/csi-driver-shared-resource-operator#24](https://github.com/openshift/csi-driver-shared-resource-operator/pull/24) not being merged yet ?

minimally yes I agree you need the csi driver hostpath volume mounts "correct" in order to employ the approach Jan provided to us

You might want to start up a local cluster, the follow the instructions at https://github.com/openshift/csi-driver-shared-resource-operator#quick-start to try out this PR and your operator PR in tandem, once you have fixed the operator PR (see my analysis of your e2e results there).

Of course, you can choose to use clusterbot as well to launch a cluster with both PRs

All this should provide you some practice in looking at must gather from our e2e runs in CI :-)

gabemontero:
/close

fixed with #77

PR #71, opened 2021-11-05
maintain secret/configmap watches/informers on demand as shares come and go; allow no resource refresh on per volume basis

gabemontero:
@openshift/openshift-team-build-api - this PR changes how we watch configmaps/secrets across namespaaces.  Where rather than watch all namespaces, but excluded some system namespaces (where that list is configurable with a series of stories from @otaviof), I create the secret/configmap shared informer related artifacts as shares are defined and recognized by this controller/driver, where we inspect the share to see what the specific namespace of the shared item is.

On share delete, we stop / prune a given namespace's shared informers if no more shares exist that point to that namespace; at the moment if is totally watch/relist event driven ... a later iteration may add an addition background monitoring loop (curious what folks think here)

but with a change like this, our scaling footprint/concerns out of the box are in many ways mitigated if this approach holds up, as we are only watching namespaces for whatever shares are created (if a user creates 1000s of shares across 1000s of namespaces, we can then tell them "don't do that" or "give our driver a bunch of extra memory" or some such)

and the reality is we do not envision such numbers at this time

additionally, if we move forward with this, tweaking the existing config probably makes sense
- don't need a  list of "ignored" namespaces

### DIFF
--- a/pkg/cache/shares.go
+++ b/pkg/cache/shares.go
@@ -1,12 +1,13 @@
 package cache
 
 import (
-	"github.com/openshift/csi-driver-shared-resource/pkg/client"
 	"sync"
 
-	"k8s.io/klog/v2"
-
 	sharev1alpha1 "github.com/openshift/api/sharedresource/v1alpha1"
+	"github.com/openshift/csi-driver-shared-resource/pkg/client"
+
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/klog/v2"
 )
 
 /*
@@ -94,6 +95,28 @@ func UpdateSharedConfigMap(share *sharev1alpha1.SharedConfigMap) {
 func UpdateSharedSecret(share *sharev1alpha1.SharedSecret) {
 	klog.V(4).Infof("UpdateSharedSecret key %s", share.Name)
 	AddSharedSecret(share)
+}
+
+func NamespacesWithSharedConfigMaps() map[string]struct{} {
+	namespacesMap := map[string]struct{}{}
+	list, err := client.GetListers().SharedConfigMaps.List(labels.Everything())
+	if err == nil {
+		for _, scm := range list {
+			namespacesMap[scm.Spec.ConfigMapRef.Namespace] = struct{}{}
+		}
+	}
+	return namespacesMap
+}
+
+func NamespacesWithSharedSecrets() map[string]struct{} {
+	namespacesMap := map[string]struct{}{}
+	list, err := client.GetListers().SharedSecrets.List(labels.Everything())
+	if err == nil {
+		for _, ss := range list {
+			namespacesMap[ss.Spec.SecretRef.Namespace] = struct{}{}
+		}
+	}
+	return namespacesMap
 }
 
 // DelSharedConfigMap removes the SharedConfigMap from our various tracking maps and calls the registered callbacks
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -2,14 +2,12 @@ package controller
 
 import (
 	"fmt"
-	"strings"
+	"sync"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/informers"
-	"k8s.io/client-go/informers/internalinterfaces"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
@@ -18,7 +16,6 @@ import (
 	sharev1alpha1 "github.com/openshift/api/sharedresource/v1alpha1"
 	objcache "github.com/openshift/csi-driver-shared-resource/pkg/cache"
 	"github.com/openshift/csi-driver-shared-resource/pkg/client"
-	"github.com/openshift/csi-driver-shared-resource/pkg/hostpath"
 	"github.com/openshift/csi-driver-shared-resource/pkg/metrics"
 
 	shareinformer "github.com/openshift/client-go/sharedresource/informers/externalversions"
@@ -28,6 +25,12 @@ const (
 	DefaultResyncDuration = 10 * time.Minute
 )
 
+type sharedObjectInformer struct {
+	stopCh          chan struct{}
+	informerFactory informers.SharedInformerFactory
+	informer        cache.SharedIndexInformer
+}
+
 type Controller struct {
 	kubeClient kubernetes.Interface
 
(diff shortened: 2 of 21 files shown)


gabemontero:
a nil ptr / panic when refresh resources is false ... otherwise doesn't look bad e2e wise

I'll debug / fix the refresh resources == false on Monday

gabemontero:
ok no refresh resources is passing 

about to push some additional changes

gabemontero:
/assign @coreydaley 

I may still add a new commit to this for turning off refresh at the volume declaration level (or I may do it in a follow up PR)

But if you have cycles, please go ahead and provide feedback for the changes here.

I'm also curious on your take for an additional background monitoring loop for cleaning up secret/configmap shareinformers.

pkg/controller/controller.go
```diff
@@ -512,18 +579,37 @@ func (c *Controller) syncSharedConfigMap(event client.Event) error {
...
 	klog.V(4).Infof("verb %s share name %s", event.Verb, share.Name)
+	var err error
 	switch event.Verb {
 	case client.DeleteObjectAction:
-		objcache.DelSharedConfigMap(share)
+		ns := objcache.DelSharedConfigMap(share)
+		if c.refreshResources {
+			c.PruneConfigMapInformers(ns)
+
+		}
```
coreydaley, line 589: This comment applies to all of the functions that you updated to return the namespaces.

I think you would be better off just creating a method that can be called from here to get the list of namespaces, instead of returning them from each function.

```suggestion
		objcache.DelSharedConfigMap(share)
		if c.refreshResources {
			c.PruneConfigMapInformers(NamespacesWithSharedConfigMaps())

		}
```

coreydaley:
What do you think of using a Synchronized Map to keep a list of the namespaces and how many shared secrets or configmaps it has? Increment during an add, decrement during a delete, leave it alone during an update. Or something similar.

gabemontero:
I'm going to cycle back to this today @coreydaley, pulling in your suggestions, but as an FYI, I am putting a 

/hold

on this as well, as we need to merge #77 first.  I'll then rebase this on top of that.

I believe I also have a commit in my local branch which I have not pushed yet that toggles resource refresh on a per share basis.  If I am correct, I will probably go ahead and push that up for review as well.

gabemontero:
OK @coreydaley I've updated with your idea ... aside from liking it in the context of this PR, it helped me on a path to rebase on top of #77, as I reworked the namespace list functions to use the listers vs. the proprietary "objcache", which I'm nuking in #77 as "unnecessary function"

Also pushed the commit that allows for disabling resource refresh at the share level .... this will help openshift build consumption of shared resources

gabemontero:
and the e2e's are still green ... very good

pkg/cache/shares.go
```diff
@@ -160,11 +165,42 @@ func UpdateSharedSecret(share *sharev1alpha1.SharedSecret) {
 	br := share.Spec.SecretRef
 	key := BuildKey(br.Namespace, br.Name)
 	secretsWithShare.Delete(key)
-	AddSharedSecret(share)
+	return
+}
+
+func NamespacesWithSharedConfigMaps() []string {
+	namespacesMap := map[string]bool{}
```
coreydaley, line 172:
```suggestion
	namespacesMap := map[string]struct{}
```

You can save some memory here by using an empty struct instead of a bool, then just do 
```
			namespacesMap[scm.Spec.ConfigMapRef.Namespace] = struct{}{}
```

Same thing with `NamespacesWithSharedSecrets` below.

coreydaley (changes requested):

pkg/cache/shares.go
gabemontero: updated @coreydaley

pkg/cache/shares.go
```diff
@@ -160,11 +165,42 @@ func UpdateSharedSecret(share *sharev1alpha1.SharedSecret) {
...
+	namespaces := []string{}
+	list, err := client.GetListers().SharedConfigMaps.List(labels.Everything())
+	if err == nil {
+		for _, scm := range list {
+			namespacesMap[scm.Spec.ConfigMapRef.Namespace] = struct{}{}
+		}
+	}
+	for ns := range namespacesMap {
+		namespaces = append(namespaces, ns)
+	}
```
coreydaley, line 182: I would contend that it would be better to refactor the `PruneSecretInformers` function to take a `map[string]struct{}` as a parameter, and get rid of translating the `namespacesMap` to a `[]string` here and returning it. It would save some processor cycles and memory.

coreydaley (changes requested):

pkg/cache/shares.go
gabemontero: agreed ... will update

gabemontero: updated @coreydaley

pkg/cache/shares.go
```diff
@@ -160,11 +165,34 @@ func UpdateSharedSecret(share *sharev1alpha1.SharedSecret) {
 	br := share.Spec.SecretRef
 	key := BuildKey(br.Namespace, br.Name)
 	secretsWithShare.Delete(key)
-	AddSharedSecret(share)
+	return
+}
+
+func NamespacesWithSharedConfigMaps() map[string]struct{} {
+	namespacesMap := map[string]struct{}{}
```
coreydaley, line 172:
```suggestion
	namespacesMap := make(map[string]struct{})
```

pkg/cache/shares.go
```diff
@@ -160,11 +165,34 @@ func UpdateSharedSecret(share *sharev1alpha1.SharedSecret) {
...
+	if err == nil {
+		for _, scm := range list {
+			namespacesMap[scm.Spec.ConfigMapRef.Namespace] = struct{}{}
+		}
+	}
+	return namespacesMap
+}
+
+func NamespacesWithSharedSecrets() map[string]struct{} {
+	namespacesMap := map[string]struct{}{}
```
coreydaley, line 183:
```suggestion
	namespacesMap := make(map[string]struct{})
```

pkg/cache/shares.go
gabemontero: why?

in the past in OCP what I did was explicitly preferred over using `make`

coreydaley: That's fine, i thought `make` was the golang gold standard.

gabemontero:
/assign @prietyc123

for qe approved label

/assign @rolfedh

for docs approved label

/assign @adambkaplan

for px approved label

rolfedh:
@gabemontero Do these changes have any doc impacts?

gabemontero:
> @gabemontero Do these changes have any doc impacts?

one of the commits here is user facing @rolfedh ... I've captured the simple description for it in our google doc rough draft of the documentation at https://docs.google.com/document/d/1F3VFkiLUJbsUkFJcjnA2oyROYTI5RRlTJYl4ZWQoNls/edit#heading=h.mqxb1y6xqdxf

rolfedh (approved):

gabemontero:
Priti is out sick

/assign @jitendar-singh

for qe approved

gabemontero:
FYI I've started the rebase for this PR after #77 merged

as expected, the rebase is "non trivial" ;-)

gabemontero:
/hold cancel

ok @coreydaley and other assigned folks on this ... I've completed the rebase after #77 merged

minimally, hold off on reviews until the e2e's have completed and are green.  I'll report status back here once they have finished

gabemontero:
yeah the e2e's are not passing ... I'll need to debug 

/hold

gabemontero:
interesting the trickier disruptive/slow tests passed, and the presumably "simpler" base line e2e is the one with the failures

so hopefully the issue is some functional edge case thing which is easy to fix

gabemontero:
hmmm .... I ran the base e2e against my personal cluster for today with the image from this PR deployed, and it as all green

was the prior failure just a flake?

/test /test e2e-aws-csi-driver

gabemontero:
sure enough ... passed now too

looking at the failure again, the tests that failed seemed to have constraints and test namespace creation was too slow and that is why our test exited ... so I think it was an actual "env flake"

releasing the hold, and asking for code review to commence (you good with picking back up from where you left off before the rebase @coreydaley ?)

/hold cancel

also going to run the test again for warm fuzzies

/test e2e-aws-csi-driver

PR #72, opened 2021-11-08
Port offset for the metrics server test

alicerum:
This change introduces port offset for the metrics server test, so that
tests use different port in case they are run in parallel and no race
condition happens.


gabemontero:
go ahead and port https://github.com/openshift/csi-driver-shared-resource-operator/pull/30 here at your convenience @alicerum - thanks !

PR #77, opened 2021-11-19
BUILD-382: fix umount issue, cross volume coruption issue; complete read only on restart support

gabemontero:
OK, this PR  started as helping @akram with #63 and his restart driver support for read only volumes, but broadened from there.  The first 6 commits of this PR are a vendor bump and either direct pull of commits from #63 or minor tweaks to commits there.

@akram - if you have cycles Thursday/Friday while @coreydaley and I are out for Thanksgiving, please take a look.  Otherwise, let's all three of us review this when we are back next week.

First, two primary bugs were addressed in getting things working:
1) use of `mount.CleanupMountPoint` from k8s did not sufficiently recognize mounts with arbitrary strings as identifiers (an approach we are employing on advice from Jan from the storage team), and calls to `umount` never occurred.  This lead to shutdowns for Pod's with our volume to hang.  We now employ `mount.Umount` directly, and shutdowns now work.  The removal of files are handled by both our code outside of this call, as well as the fact that we employ tmpfs, and once the unmount occurs, the files disappear.

this fix is encapsulated in commit https://github.com/openshift/csi-driver-shared-resource/pull/77/commits/61e3108225463eec34dbdfcaaf73f818db0145dc

2) the storage of our volume state to disk as a single file of encoded `map[string]hostPathVolume{}` suffered from intermittent concurrency issues, where the value for one key would be stored under another key; after several days of debug, I became pretty certain the issue was not our driver storing the wrong value under the wrong key, but something in the golang json encoding level.  This has since been proven with the switch to a file per volume, were the file name is the same as the volume ID, as that was the key we are previously using to store entries in that map.  The intermittent errors in our disruptive e2e went away, as confirmed by repeated testing once this change was made (where previously, I could reproduced the problem after an iteration or two after 2 to 6 pods were created.

this fix is encapsulated as part of the changes in commit https://github.com/openshift/csi-driver-shared-resource/pull/77/commits/e3f8d7502f403bdb62934529798bc90c3da2aa1d

3) prior to figuring out 2) ^^, I considered our 'objcache' a candidate for the intermittent bug, and ended up removing a bunch of this code, and just use our listers / shared informers from the controller.  This ultimately did not fix 2), but the change is really an improvement (including deleting a bunch of code).  So I'm keeping it here.

4) As part of doing 3), I consolidated our use of listers when we refresh resources and the k8s client when we do not refresh resources, into single methods, and tweaked how we check for "not refreshing resources" by working of the config value vs. the presence of the client reference.  Again, did not fix 2), but looks way better IMO, so kept it.

5) lastly, reworked the inline functions for the various sync map range callbacks a bit, and associated them with structs with variables.  Again, did not fix 2), but again felt "cleaner", so kept it.


gabemontero:
/assign @coreydaley 
/assign @akram 

please see the updated PR description for full context

@akram if you have cycles to look at while @coreydaley and I are on PTO, great.  Otherwise, we can tackle this next week when we are all on the clock.

cmd/main.go
```diff
@@ -152,6 +167,7 @@ func setupSignalHandler() (stopCh <-chan struct{}) {
 		close(stop)
 		<-c
 		os.Exit(1) // second signal. Exit directly.
+		klog.Info("exiting")
```
coreydaley, line 170: `exiting` may not make much sense when viewed in a log file without any context. Maybe a bit more information here would be useful.

pkg/cache/configmaps.go
```diff
@@ -39,93 +33,50 @@ var (
...
+			if share.Spec.ConfigMapRef.Namespace == configmap.Namespace && share.Spec.ConfigMapRef.Name == configmap.Name {
+				shareSecretsUpdateCallbacks.Range(buildRanger(buildCallbackMap(share.Name, share)))
+			}
+		}
+	} else {
+		klog.Warningf("error during UpsertConfigMap on shared configmaps lister list: %s", err.Error())
 	}
 	// otherwise process any share that arrived after the configmap
-	configmapsWithShares.Store(key, configmap)
+	//configmapsWithShares.Store(key, configmap)
```
coreydaley, line 57:
```suggestion
	
```

pkg/cache/configmaps.go
```diff
@@ -39,93 +33,50 @@ var (
...
+	sharecConfigMapList, err := client.GetListers().SharedConfigMaps.List(labels.Everything())
+	if err == nil {
+		for _, share := range sharecConfigMapList {
+			if share.Spec.ConfigMapRef.Namespace == configmap.Namespace && share.Spec.ConfigMapRef.Name == configmap.Name {
+				shareSecretsUpdateCallbacks.Range(buildRanger(buildCallbackMap(share.Name, share)))
+			}
+		}
+	} else {
+		klog.Warningf("error during UpsertConfigMap on shared configmaps lister list: %s", err.Error())
 	}
```
coreydaley, line 55:
```suggestion
	shareConfigMapList, err := client.GetListers().SharedConfigMaps.List(labels.Everything())
	if err != nil {
		klog.Warningf("error during UpsertConfigMap on shared configmaps lister list: %s", err.Error())
	}
	for _, share := range shareConfigMapList {
		if share.Spec.ConfigMapRef.Namespace == configmap.Namespace && share.Spec.ConfigMapRef.Name == configmap.Name {
			shareSecretsUpdateCallbacks.Range(buildRanger(buildCallbackMap(share.Name, share)))
		}
	}		
	
```

pkg/cache/secrets.go
```diff
@@ -40,93 +34,50 @@ var (
...
+	sharedSecretsList, err := client.GetListers().SharedSecrets.List(labels.Everything())
+	if err == nil {
+		for _, share := range sharedSecretsList {
+			if share.Spec.SecretRef.Namespace == secret.Namespace && share.Spec.SecretRef.Name == secret.Name {
+				shareSecretsUpdateCallbacks.Range(buildRanger(buildCallbackMap(share.Name, share)))
+			}
+		}
+	} else {
+		klog.Warningf("error during UpsertSecret on shared secrets lister list: %s", err.Error())
 	}
```
coreydaley, line 56:
```suggestion
	sharedSecretsList, err := client.GetListers().SharedSecrets.List(labels.Everything())
	if err != nil {
		klog.Warningf("error during UpsertSecret on shared secrets lister list: %s", err.Error())
	}
	for _, share := range sharedSecretsList {
		if share.Spec.SecretRef.Namespace == secret.Namespace && share.Spec.SecretRef.Name == secret.Name {
			shareSecretsUpdateCallbacks.Range(buildRanger(buildCallbackMap(share.Name, share)))
		}
	}
	
```

pkg/cache/shares.go
```diff
@@ -84,82 +75,27 @@ func AddSharedSecret(share *sharev1alpha1.SharedSecret) {
 	br := share.Spec.SecretRef
 	key := BuildKey(br.Namespace, br.Name)
 	klog.V(4).Infof("AddSharedSecret key %s", key)
-	obj, ok := secrets.Load(key)
-	if obj != nil && ok {
-		s := obj.(*corev1.Secret)
-		secretsWithShare.Store(key, s)
+	//obj, ok := secrets.Load(key)
```
coreydaley, line 78:
```suggestion

```

pkg/cache/shares.go
```diff
@@ -63,19 +58,15 @@ var (
...
 		// so this line build a map with a single entry, the share from this event, and then
 		// applies the function(s) supplied by the CSI volume code in order to make changes based
 		// on this event
+		klog.V(4).Infof("AddSharedConfigMap share %s key %s start range", share.Name, key)
 		shareConfigMapsUpdateCallbacks.Range(buildRanger(buildCallbackMap(share.Name, share)))
-		//NOTE we do not store share in shareConfigMaps unless the backing resource is available
-		shareConfigMaps.Store(share.Name, share)
-	} else {
-		sharesWaitingOnConfigmaps.Store(share.Name, share)
+		klog.V(4).Infof("AddSharedConfigMap share %s key %s end range", share.Name, key)
```
coreydaley, line 69: This looks like the same message that is logged if it succeeds. Should this be a `Warning` at least and a different message?

pkg/cache/shares.go
```diff
@@ -84,82 +75,27 @@ func AddSharedSecret(share *sharev1alpha1.SharedSecret) {
 	br := share.Spec.SecretRef
 	key := BuildKey(br.Namespace, br.Name)
 	klog.V(4).Infof("AddSharedSecret key %s", key)
-	obj, ok := secrets.Load(key)
-	if obj != nil && ok {
-		s := obj.(*corev1.Secret)
-		secretsWithShare.Store(key, s)
+	//obj, ok := secrets.Load(key)
+	s := client.GetSecret(br.Namespace, br.Name)
+	if s != nil {
```
coreydaley, line 80: Should there be an `else` that logs at least a `Warning` if this fails?

pkg/config/manager.go
```diff
@@ -60,6 +60,7 @@ func (m *Manager) LoadConfig() (*Config, error) {
 	if err = yaml.Unmarshal(payload, &cfg); err != nil {
 		return nil, err
 	}
+	LoadedConfig = cfg
```
coreydaley, line 63: I don't see `LoadedConfig` being used anywhere else besides here and being declared in `pkg/config/config.go`. Was this leftover from something else?

coreydaley (changes requested):

cmd/main.go
gabemontero: this ended up not proving useful ... I'm just going to remove, thanks.

pkg/cache/shares.go
gabemontero: this is some left over debug bracket tracing .... don't need it I'll just remove, thanks..

pkg/cache/shares.go
gabemontero: yeah a share pointing to a  secret / configmap that does not exist is a "user error" that could emerge, where we don't want repeated error logging for it

the situation is easily discernable, in that the volume mount will not have any content, and the user can look up the share, then try to look up the shared resource and see not found

Still though ... I'll compromise and add a level 4 trace in the client.GetSecret / client.GetConfigMap calls

pkg/config/manager.go
gabemontero: It is used in `NewHostPathDriver` and `Run` in hostpath.go

gabemontero: Perhaps the use was in a separate commit from the one that has this reference (a guess, did not investigate to confirm)

gabemontero:
pushed updates in separate commit @coreydaley  ... also found some more commented out code to delete while addressing you items

pkg/config/manager.go
coreydaley: I loaded all commits and used Ctrl-F in the browser. Not a big deal.

gabemontero:
aws fail on disruptive e2e (whew)

/test e2e-aws-csi-driver-disruptive

rolfedh (commented):
@gabemontero Do these changes have any doc impacts?

gabemontero:
> @gabemontero Do these changes have any doc impacts?

These changes do no @rolfedh  ..... they address known limitations in shared resources which we did not want to have to document, and now do not.

rolfedh (approved):

gabemontero:
Priti is out sick

/assign @jitendar-singh

for qe approved

gabemontero:
uh oh .... did not catch that @rolfedh had applied the lgtm label @coreydaley as part of trying to "docs-approve" this  :-/

if you still have pending comments, do still post them here, and I'll craft a new PR  to incorporate them.

rolfedh:
Sorry about that, @gabemontero and @coreydaley. I'll use only the `/label docs-approved` next time.

PR #78, opened 2021-11-24
metrics: split openshift_csi_share_mount_total into total and failures

jan--f:
As proposed in https://github.com/openshift/cluster-monitoring-operator/pull/1477 this spits the metric `openshift_csi_share_mount_total` into `openshift_csi_share_mount_requests_total` and `openshift_csi_share_mount_failures_total`.

Feel free to stick with what is in the code now. I wanted to propose this as I approved the current state in #54.

PR #79, opened 2021-11-29
Split and rename sci share mount total metric

alicerum:
Monitoring team recommends to split `success`/`fail` labels into different metrics and call them `fail_total` and `requests_total` in order to be able to access failed percentage easily.

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -1,8 +1,6 @@
 package metrics
 
 import (
-	"strconv"
-
 	"github.com/prometheus/client_golang/prometheus"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/metrics/metrics.go
```diff
@@ -11,28 +9,43 @@ const (
 
 	sharesSubsystem = "openshift_csi_share"
 
-	mount          = "mount"
-	mountCountName = sharesSubsystem + separator + mount + separator + "total"
+	mount  = "mount"
+	failed = "failed"
+
+	mountCountName  = sharesSubsystem + separator + mount + separator + "total"
```
gabemontero, line 15: I think the `requests_total` and `failure_totals` names that @jan--f has in #78 is better for these @alicerum

gabemontero (changes requested):
/approve

but I'm asking for 1 reconciliation between what you have @alicerum and what @jan--f has in #78

pkg/metrics/metrics.go
gabemontero: ok assign yourself to that one @alicerum and give it an lgtm .... I'll then approve

PR #81, opened 2021-12-08
Support TLS in the metrics endpoint

alicerum:
Metrics endpoint must now support TLS provided in the secret mounted by the CSI SR operator.

### DIFF
--- a/pkg/metrics/server.go
+++ b/pkg/metrics/server.go
@@ -11,6 +11,14 @@ import (
 	"k8s.io/klog/v2"
 )
 
+var (
+	// these files are mounted from the openshift secret
+	// shared-resource-csi-driver-node-metrics-serving-cert
+	// by the csi-driver-shared-resource-operator
+	tlsCRT = "/etc/secrets/tls.crt"
+	tlsKey = "/etc/secrets/tls.key"
+)
+
 // BuildServer creates the http.Server struct
 func BuildServer(port int) (*http.Server, error) {
 	if port <= 0 {
@@ -41,7 +49,7 @@ func StopServer(srv *http.Server) {
 // RunServer starts the metrics server.
 func RunServer(srv *http.Server, stopCh <-chan struct{}) {
 	go func() {
-		err := srv.ListenAndServe()
+		err := srv.ListenAndServeTLS(tlsCRT, tlsKey)
 		if err != nil && err != http.ErrServerClosed {
 			klog.Errorf("error starting metrics server: %v", err)
 		}
--- a/pkg/metrics/server_test.go
+++ b/pkg/metrics/server_test.go
@@ -1,8 +1,17 @@
 package metrics
 
 import (
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/pem"
 	"fmt"
+	"io/ioutil"
+	"math/big"
+	mr "math/rand"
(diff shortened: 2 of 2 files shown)


alicerum:
/assign @prietyc123
For the qe label

and

/assign @rolfedh
for the docs approve label

alicerum:
/assign @gabemontero
for Gabe's approval label

alicerum:
For this PR to work correctly we need to first get https://github.com/openshift/csi-driver-shared-resource-operator/pull/37 merged, as node metric server expects TLS secrets to be mounted by the operator.

pkg/metrics/server_test.go
```diff
@@ -19,9 +28,67 @@ var (
...
+	if err != nil {
+		panic(err)
+	}
+
+	// sets the default http client to skip certificate check.
+	http.DefaultTransport.(*http.Transport).TLSClientConfig = &tls.Config{
+		InsecureSkipVerify: true,
+	}
+
+	code := m.Run()
```
gabemontero, line 46: should we be verifying that we get a healthy recurn code here ?

pkg/metrics/server_test.go
```diff
@@ -19,9 +28,67 @@ var (
...
+	os.Remove(tlsKey)
+	os.Remove(tlsCRT)
+	os.Exit(code)
```
gabemontero, line 49: curious what this achieves?  not apparent to me

if this has some value, perhaps a comment above to explain

pkg/metrics/server.go
```diff
@@ -11,6 +11,11 @@ import (
 	"k8s.io/klog/v2"
 )
 
+var (
+	tlsCRT = "/etc/secrets/tls.crt"
```
gabemontero, line 15: how did you derive these locations ... a comment here explaining that would be good

gabemontero (changes requested):
just some minor items @alicerum

pkg/metrics/server.go
alicerum: this is the location to which I mounted the secret volume in the https://github.com/openshift/jenkins-sync-plugin/pull/408 PR

gabemontero: huh .... was does a jenkins sync plugin change have to do with shared resource csi driver?

alicerum: @gabemontero of course i meant my CSO PR: https://github.com/openshift/cluster-storage-operator/pull/243

gabemontero: ok that make sense :-)

now add a comment in the source here referring to that mount being done in the driver's deployment (which is managed by the CSO)

gabemontero:
ran into the RH registry image pulling delay I'm addressing in #82 

/retest

gabemontero:
terraform error 

/retest

gabemontero:
bump @prietyc123  - are you able to asses this PR for applying the qe approved label?

pkg/metrics/server_test.go
alicerum: this is a TestMain method that sets up tlsKey and tlsCert for all the other test methods. We need to do this in order to create server with correct testing tls keys for each test.
TestMain method, if exists, is the main entrypoint for the unit tests. It should return the exit code of `m.Run()`, which runs all the other test methods, and it allows to set common state for the testing environment.
I can add it to the comments if needed.

PR #84, opened 2022-01-14
BUILD-381: require CSI Volumes using the Shared Resources driver to be specified as readOnly == true

gabemontero:
Also fixed a doc update that was missed when I changed the controller watch to listen to the namespaces that are referenced by shares instead of listening to all namespaces but filter out some openshift system namespaces.

/assign @coreydaley 
a lot of deleted code 
I would also use `?w=1` at the end of the git URL used during code review in your browser to filter down white space differences 

for PR review

@akram FYI - yeah a fair amount of e2e reduction was needed since we can no longer support shares mounted off other shares since we require read only to be true

### DIFF
--- a/docs/content-update-details.md
+++ b/docs/content-update-details.md
@@ -1,43 +1,50 @@
 # Details around pushing Secret and ConfigMap updates to provisioned Volumes
 
-### Excluded OCP namespaces
+By default, the Shared Resource CSI Driver will watch for updates to `Secrets` and 
+`ConfigMaps` in any of the namespaces referenced in corresponding `SharedSecrets` and 
+`SharedConfigMaps`.  Then, when the actual `Secrets` or `ConfigMaps` referenced by those
+`SharedSecrets` and `SharedConfigMaps` change, each volume across all the active `Pods`
+in the system will have their content corresponding to those `Secrets` and `ConfigMaps`updated.
 
-The current list of namespaces excluded by default from the controller's watches:
+NOTE: this driver can still update the contents of the `Volumes` it provisions, even as it 
+requires `Pods` to mark the `Volumes` as read-only.
 
-- kube-system
-- openshift-machine-api
-- openshift-kube-apiserver
-- openshift-kube-apiserver-operator
-- openshift-kube-scheduler
-- openshift-kube-controller-manager
-- openshift-kube-controller-manager-operator
-- openshift-kube-scheduler-operator
-- openshift-console-operator
-- openshift-controller-manager
-- openshift-controller-manager-operator
-- openshift-cloud-credential-operator
-- openshift-authentication-operator
-- openshift-service-ca
-- openshift-kube-storage-version-migrator-operator
-- openshift-config-operator
-- openshift-etcd-operator
-- openshift-apiserver-operator
-- openshift-cluster-csi-drivers
-- openshift-cluster-storage-operator
-- openshift-cluster-version
-- openshift-image-registry
-- openshift-machine-config-operator
-- openshift-sdn
-- openshift-service-ca-operator
+This default behavior can be disabled at both [global level](config.md) and a per volume level.
 
-The list is configurable by informing `--ignorenamespace` to the `hostpath` plugin instance. The
-plugin can also be configured with `--refreshresources` flag, which makes the controller keep a warm
-cache of `ConfigMap` and `Secret` objects, and as they change the controller will follow the updates.
+The global setting takes precedence.  It is considered the domain of the cluster administrator,
+with the belief that if they want resource refreshing disabled, it should be disabled for everyone.
 
-When `--refreshresources` is disabled (i.e. `--refreshresources=false`), the controller will read the
-backing-resource (`.spec.backingResource`) just before mounting the volume instead of keeping a warm
-cache. Additionally, every volume mount will be always read-only preventing tampering with the data
-provided by this CSI driver.
+For disabling at the volume specific level, the `volumeAttributes` field should have an entry with the 
+key `refreshResource` and a value of `false`.
 
-Allowing the disabling processing of updates, or switching the default for the system as not dealing with
-updates, but then allowing for opting into updates, is also under consideration.
+```yaml
+kind: Pod
+apiVersion: v1
+metadata:
+  name: my-csi-app
+  namespace: my-csi-app-namespace
+spec:
+  serviceAccountName: default
+  containers:
+  
+  # specific container content would be here
+  
+  
+  volumes:
+    - name: my-csi-volume
+      csi:
+        readOnly: true
+        driver: csi.sharedresource.openshift.io
+        volumeAttributes:
+          sharedConfigMap: my-share
+          refreshResource: false
+```
+
+## NOTES
+
+1) If the inverse on order of precedence gains favor as users start using this driver in earnest, we'll 
+look into ways of providing that pattern.
+
+2) This repository's maintainers are well aware of the "atomic writer" concept in upstream Kubernetes for coordinating
+a Pod's access to secret volumes with the system's attempt to update them.  Such support is not yet integrated with
+this driver, but we have plans to do so in a future release.
--- a/docs/csi.md
+++ b/docs/csi.md
@@ -6,11 +6,11 @@ So let's take each part of the [CSIVolumeSource](https://github.com/kubernetes/a
 - for the `VolumeAttributes` map, this driver currently inspects the "sharedConfigMap" key or "sharedSecret" key (which map the `SharedConfigMap` OR `SharedSecret` instance your `Pod` wants to use) in addition to the
   elements of the `Pod` the kubelet stores when contacting the driver to provision the `Volume`.  See [this list](https://github.com/openshift/csi-driver-shared-resource/blob/c3f1c454f92203f4b406dabe8dd460782cac1d03/pkg/hostpath/nodeserver.go#L37-L42).
 - NOTE: you cannot specify both a "sharedConfigMap" and "sharedSecret" key.  An error will be flagged.  An error will also be flagged if neither is present, or if the value for one or the other does not equal the name of a `SharedConfigMap` or `SharedSecret`
-- the `ReadOnly` field is honored.  The driver can even update the content for a read-only `Volume`, even as the `Pod` consuming the `Volume` cannot write to the `Volume`.  However,
-  a limitation exists which makes any updates of content for read-only `Volumes` *OFFICIALLY* unsupported.  Namely, the function will not work across driver restarts.  The driver
-  loses access to the kubelet's mount for the `Volume`.  This is not the case for read-write `Volumes`.  The driver is able to update the content of read-write `Volumes` provisioned before the 
-  latest restart of the driver.
-- Also, mounting of one `SharedConfigMap` OR `SharedSecret` off of a subdirectory of another `SharedConfigMap` OR `SharedSecret` is only supported with read-write `Volumes`.  
+- the `ReadOnly` field is required to be set to 'true'.  This follows conventions introduced in upstream Kubernetes CSI Drivers to facilitate proper SELinux labelling.  What occurs is that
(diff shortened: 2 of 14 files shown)


gabemontero:
by code review URL I mean https://github.com/openshift/csi-driver-shared-resource/pull/84/files?w=1 @coreydaley

gabemontero:
/assign @rolfedh 

for docs approve label
I previously updated https://docs.google.com/document/d/1F3VFkiLUJbsUkFJcjnA2oyROYTI5RRlTJYl4ZWQoNls/edit#heading=h.zdumv4gow9cw
to account for these changes

gabemontero:
/assign @prietyc123 

for qe approved label

gabemontero:
> by code review URL I mean https://github.com/openshift/csi-driver-shared-resource/pull/84/files?w=1 @coreydaley

or something similar for each commit if you want to break it up that way @coreydaley

gabemontero:
aws /install fail on https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_csi-driver-shared-resource/84/pull-ci-openshift-csi-driver-shared-resource-master-e2e-aws-csi-driver-slow/1482035631273021440

fyi it passed for me locally

/retest

gabemontero:
more aws pain

/retest

docs/content-update-details.md
```diff
@@ -1,43 +1,50 @@
...
-Allowing the disabling processing of updates, or switching the default for the system as not dealing with
-updates, but then allowing for opting into updates, is also under consideration.
\ No newline at end of file
+By default, the Shared Resource CSI Driver will watch for updates to `Secrets` and 
+`ConfigMaps` in any of the namespaces referenced in corresponding `SharedSecrets` and 
+`SharedConfigMaps`.  Then, when the actual `Secrets` or `ConfigMaps` referenced by those
+`SharedSecrets` and `SharedConfigMaps` change, each volume across all the active `Pods`
+in the system will have their content corresponding to those `Secrets` and `ConfigMaps`.
+
+So while the driver requires all volume references in the various `Pods` are read only, so that
```
akram, line 9: Can we rephrase this one to be more understable?

docs/csi.md
```diff
@@ -6,11 +6,11 @@ So let's take each part of the [CSIVolumeSource](https://github.com/kubernetes/a
...
-  a limitation exists which makes any updates of content for read-only `Volumes` *OFFICIALLY* unsupported.  Namely, the function will not work across driver restarts.  The driver
-  loses access to the kubelet's mount for the `Volume`.  This is not the case for read-write `Volumes`.  The driver is able to update the content of read-write `Volumes` provisioned before the 
-  latest restart of the driver.
-- Also, mounting of one `SharedConfigMap` OR `SharedSecret` off of a subdirectory of another `SharedConfigMap` OR `SharedSecret` is only supported with read-write `Volumes`.  
+- the `ReadOnly` field is required to be set to 'true'.  This follows conventions introduced in upstream Kubernetes CSI Drivers to facilitate proper SELinux labelling.  What occurs is that
+this driver will return a read-write linux file system to the kubelet, so that CRI-O can apply the correct SELinux labels on the file system (CRI-O would not be able to update the SELinux labels on a read only file system
+after it is created), but the kubelet still makes sure that the file system later exposed to the consuming (which sits on top of the file system returned by this repository's driver) is read only.
```
akram, line 11: s/consuming /consuming pod/

pkg/hostpath/hostpath_test.go
```diff
@@ -95,318 +95,282 @@ func TestCreateHostPathVolumeBadAccessType(t *testing.T) {
 }
 
 func TestCreateDeleteConfigMap(t *testing.T) {
-	readOnly := []bool{true, false}
```
akram, line 98: in my attempt to change these tests, I was wondering if it wouldn't be interresting to keep this for loop with 2 attempts; so we are sure that a creation then a deletion then a re-creation works well.
That would change the for loop with an i++ counter instead, but that would keep the test valid and valuable; wdyt ?

pkg/hostpath/hostpath.go
```diff
@@ -50,6 +50,8 @@ type hostPath struct {
 	ephemeral         bool
 	maxVolumesPerNode int64
 
+	testSoNoMounter bool
```
akram, line 53: Shouldn't we mock/stub this interface instead of adding this parameter especially to a public method?
Also, by reading the tests comments, it seems that we will be needing to have sudo privilege to run the tests; which is a bit complex.

pkg/hostpath/hostpath.go
```diff
@@ -888,11 +835,13 @@ func (hp *hostPath) Prune(kubeClient kubernetes.Interface) {
 			continue
 		}
 		klog.V(2).Infof("pruner: removed volume file %q with missing pod from disk", file)
-		err = mounter.Unmount(hpv.GetVolPathAnchorDir())
-		if err != nil {
-			klog.V(2).Infof("pruner: issue unmounting for volume %s mount id %s: %s", hpv.GetVolID(), hpv.GetVolPathAnchorDir(), err.Error())
-		} else {
-			klog.V(2).Infof("pruner: successfully unmounted volume %s mount id %s", hpv.GetVolID(), hpv.GetVolPathAnchorDir())
+		if mounter != nil {
```
akram, line 838: so this is only needed when we are in unit tests?
The mounter here is only instantiated if we are in test mode. And so is the clean-up. I would suggest that we add this logic only in unit tests methods only.

```
err = mounter.Unmount(hpv.GetVolPathAnchorDir())
``` 
we only need a reference to the `hpv` and we could create a `mounter` in the unit test.
wdyt ?

docs/content-update-details.md
gabemontero: yep part of next push

pkg/hostpath/hostpath_test.go
gabemontero: Sorry no, I don't think such a permutation has value for any of these tests.  The volume won't even be provisioned when the kubelet asks for it if readOnly is not set to true with these changes.  That is the requirement we have had to establish, the pattern from upstream we are following, in order to get the SELinux labels correct.

pkg/hostpath/hostpath.go
gabemontero: see my comment above ... I'll create a mock/test version of the mount interface

pkg/hostpath/hostpath.go
gabemontero: fyi k8s over at https://github.com/kubernetes/utils/tree/master/mount does not have a "fake" mounter

gabemontero: and yes with all this we are "avoiding" the need for sudo priv's that are required to call the "real" mounter

gabemontero:
ok thanks for the review @akram I've either responded in comments or made change in a separate commit for this review from you.... PTAL ... I'll squash commits when we reach a steady state with your review

akram:
That looks good now @gabemontero !  You can squash then, so I can lgtm that.

pkg/hostpath/hostpath.go
akram: I am glad to see that now we have something more elegant than k8s folks :-D

gabemontero:
> That looks good now @gabemontero ! You can squash then, so I can lgtm that.

thanks @akram 

the akram-1 commit has been squashed into the code/doc/unit/e2e commits appropriately

gabemontero:
bump @prietyc123 

for qe approved label

the scenario here is to confirm that it the pod trying to mount the shared resource volume does not set the volumeAttribute `readOnly` to true, it gets a failed to mount error now, as we require that attribute to be set

docs/content-update-details.md
```diff
@@ -1,43 +1,50 @@
...
-provided by this CSI driver.
-
-Allowing the disabling processing of updates, or switching the default for the system as not dealing with
-updates, but then allowing for opting into updates, is also under consideration.
\ No newline at end of file
+By default, the Shared Resource CSI Driver will watch for updates to `Secrets` and 
+`ConfigMaps` in any of the namespaces referenced in corresponding `SharedSecrets` and 
+`SharedConfigMaps`.  Then, when the actual `Secrets` or `ConfigMaps` referenced by those
+`SharedSecrets` and `SharedConfigMaps` change, each volume across all the active `Pods`
+in the system will have their content corresponding to those `Secrets` and `ConfigMaps`.
```
coreydaley, line 7:
```suggestion
in the system will have their content corresponding to those `Secrets` and `ConfigMaps` updated.
```

docs/content-update-details.md
```diff
@@ -1,43 +1,50 @@
...
+
+NOTE: this driver can still update the contents of the `Volumes` it provisions, even as it 
+requires `Pods` to mark the `Volumes` as read-only.
+
+This default behavior can be disabled at both [global level](config.md) and a per volume level.
+
+The global setting takes precedence.  It is considered the domain of the cluster administrator,
+with the belief that if they want resource refreshing disabled, it should be disabled for everyone.
+
+For disabling at the volume specific level, the `volumeAttributes` field in should have an entry with the
```
coreydaley, line 17:
```suggestion
For disabling at the volume specific level, the `volumeAttributes` field should have an entry with the 
```

docs/csi.md
```diff
@@ -6,11 +6,11 @@ So let's take each part of the [CSIVolumeSource](https://github.com/kubernetes/a
...
-  latest restart of the driver.
-- Also, mounting of one `SharedConfigMap` OR `SharedSecret` off of a subdirectory of another `SharedConfigMap` OR `SharedSecret` is only supported with read-write `Volumes`.  
+- the `ReadOnly` field is required to be set to 'true'.  This follows conventions introduced in upstream Kubernetes CSI Drivers to facilitate proper SELinux labelling.  What occurs is that
+this driver will return a read-write linux file system to the kubelet, so that CRI-O can apply the correct SELinux labels on the file system (CRI-O would not be able to update the SELinux labels on a read only file system
+after it is created), but the kubelet still makes sure that the file system later exposed to the consuming pod (which sits on top of the file system returned by this repository's driver) is read only.
+If this driver allowed both read-only and read-write, there is in fact no way to provide differing support that still allows for correct SELinux labelling for each).
+- Also, mounting of one `SharedConfigMap` OR `SharedSecret` off of a subdirectory of another `SharedConfigMap` OR `SharedSecret` is *NOT* supported the driver only supports read-only `Volumes`.
```
coreydaley, line 13:
```suggestion
- Also, mounting of one `SharedConfigMap` OR `SharedSecret` off of a subdirectory of another `SharedConfigMap` OR `SharedSecret` is *NOT* supported. The driver only supports read-only `Volumes`.  
```

pkg/hostpath/hostpath.go
```diff
@@ -888,11 +830,13 @@ func (hp *hostPath) Prune(kubeClient kubernetes.Interface) {
...
 		klog.V(2).Infof("pruner: removed volume file %q with missing pod from disk", file)
-		err = mounter.Unmount(hpv.GetVolPathAnchorDir())
-		if err != nil {
-			klog.V(2).Infof("pruner: issue unmounting for volume %s mount id %s: %s", hpv.GetVolID(), hpv.GetVolPathAnchorDir(), err.Error())
-		} else {
-			klog.V(2).Infof("pruner: successfully unmounted volume %s mount id %s", hpv.GetVolID(), hpv.GetVolPathAnchorDir())
+		if hp.mounter != nil {
+			err = hp.mounter.Unmount(hpv.GetVolPathAnchorDir())
+			if err != nil {
+				klog.V(2).Infof("pruner: issue unmounting for volume %s mount id %s: %s", hpv.GetVolID(), hpv.GetVolPathAnchorDir(), err.Error())
```
coreydaley, line 836: I think that this should be a warning, not an info.

coreydaley (changes requested):

gabemontero:
install pain no refresh e2e

/retest

PR #85, opened 2022-01-25
add custom must-gather image for 'oc adm must-gather'

gabemontero:
See https://docs.openshift.com/container-platform/4.9/support/gathering-cluster-data.html#gathering-data-specific-features_gathering-cluster-data for how some other "features" provide additional must-gather semantics.

I patterned this after what was done in the local-storage-operator.  That feature seemed the closest / most similar to what shared resources is.

So in theory we would ultimately have an image like `registry.redhat.io/openshift4/ocp-shared-resource-mustgather-rhel8` image available.

Based on the pattern demonstrated there, it does not look like we want to pursue getting this in the payload.  However perhaps a question to ask (during the 4.11 cycle)

Next question, does that mean we still pursue things through ART, since shared resources is part of the payload, regardless of where our must-gather image lands, or is this a CPaaS item.  Based on the recent 4.11 kick off, that has me leaning toward guessing CPaaS.

That said, after this PR merges, I think we can still pursue openshift/release work for the new image, even before any payload/ART/CPaas decisions, like what occurred with shared resources.  IIRC, that at least gets us a quay.io/openshift/origin-... image that we could direct customers too in a pinch.

Given that, perhaps we pursue merging this and an openshift/release PR before 4.10 code freeze.

But let's discuss.

Finally, FYI, `shellcheck` reports no items to correct with the new bash file `gather`.  I did not add the `.sh` suffix because that is what local-storage-operator did.

@openshift/openshift-team-build-api FYI / ptal

Testing locally, running `oc adm must-gather --image=docker.io/gmontero/origin-csi-driver-shared-resource-mustgather:latest` adds the provisioned volume metadata the driver stores in its `/csi-volumes-map` hostpath volume in a subfolder under the cluster scoped resources -> shared resources folder.  That lines up with what https://github.com/openshift/cluster-storage-operator/pull/257 will create when gathering shared resources instances when base OCP must gather is run.

### DIFF
--- a/Dockerfile.mustgather
+++ b/Dockerfile.mustgather
@@ -1,0 +1,9 @@
+FROM registry.ci.openshift.org/ocp/4.10:must-gather AS builder
+WORKDIR /go/src/github.com/openshift/csi-driver-shared-resource
+COPY . .
+
+FROM registry.ci.openshift.org/ocp/4.10:cli
+COPY --from=builder /go/src/github.com/openshift/csi-driver-shared-resource/must-gather/* /usr/bin/
+RUN chmod +x /usr/bin/gather
+
+ENTRYPOINT /usr/bin/gather
--- a/Makefile
+++ b/Makefile
@@ -73,6 +73,12 @@ build-image: ## Build the images and push them to the remote registry. Example: 
 	$(CONTAINER_RUNTIME) push $(REGISTRY)/$(REPOSITORY)/origin-csi-driver-shared-resource:$(TAG)
 .PHONY: build-image
 
+build-mustgather-image: ## Build the customer must gather images and push them to the remote registry. Example: make build-mustgather-image
+	rm -rf _output
+	$(CONTAINER_RUNTIME) build -f Dockerfile.mustgather -t $(REGISTRY)/$(REPOSITORY)/origin-csi-driver-shared-resource-mustgather:$(TAG) .
+	$(CONTAINER_RUNTIME) push $(REGISTRY)/$(REPOSITORY)/origin-csi-driver-shared-resource-mustgather:$(TAG)
+.PHONY: build-mustgather-image
+
 clean: ## Clean up the workspace. Example: make clean
 	rm -rf _output
 .PHONY: clean
--- a/must-gather/gather
+++ b/must-gather/gather
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+pods=()
+
+cdsrNamespace="openshift-cluster-csi-drivers"
+for I in $(oc get pod -n ${cdsrNamespace} -o custom-columns=NAME:.metadata.name --no-headers); do
+  pods+=("$I")
+done
+
+object_collection_path=must-gather/cluster-scoped-resources/sharedresource.openshift.io
+mkdir -p ${object_collection_path}
+
+for pod in "${pods[@]}"; do
+  if [[ $pod != shared-resource-csi-driver-node* ]] ;
+  then
+    continue
+  fi
+  mkdir -p ${object_collection_path}/"$pod"
+  oc rsync -c hostpath -n ${cdsrNamespace} "$pod":/csi-volumes-map ${object_collection_path}/"$pod"
+done
+
+exit 0


Dockerfile.mustgather
```diff
@@ -0,0 +1,9 @@
+FROM registry.ci.openshift.org/ocp/4.10:must-gather AS builder
+WORKDIR /go/src/github.com/openshift/csi-driver-shared-resource
+COPY . .
+
+FROM registry.ci.openshift.org/ocp/4.10:cli
+COPY --from=builder /go/src/github.com/openshift/csi-driver-shared-resource/must-gather/* /usr/bin/
```
adambkaplan, line 6: Why do we need to run a multi-stage build? Isn't the base `cli` image sufficient, since this is just running bash?

adambkaplan:
> Based on the pattern demonstrated there, it does not look like we want to pursue getting this in the payload. However perhaps a question to ask (during the 4.11 cycle)

Are the other must-gather extensions in the payload? If there is an image reference in the cluster-storage-operator, the answer is yes.

> Next question, does that mean we still pursue things through ART, since shared resources is part of the payload, regardless of where our must-gather image lands, or is this a CPaaS item. Based on the recent 4.11 kick off, that has me leaning toward guessing CPaaS.

I think the long term strategy is to lessen the footprint of OpenShift - net new things should strive to be delivered on top with an operator. This is an extension to an item that exists in the payload and will be lifecycled with OpenShift, so I my first guess is that it will likewise go in the payload.

> That said, after this PR merges, I think we can still pursue openshift/release work for the new image, even before any payload/ART/CPaas decisions, like what occurred with shared resources. IIRC, that at least gets us a quay.io/openshift/origin-... image that we could direct customers too in a pinch.

Agree here - we'll want to build the new image in CI and mirror it to quay.io before doing anything else.

gabemontero:
> > Based on the pattern demonstrated there, it does not look like we want to pursue getting this in the payload. However perhaps a question to ask (during the 4.11 cycle)
> 
> Are the other must-gather extensions in the payload? If there is an image reference in the cluster-storage-operator, the answer is yes.

yeah only core must gather is there .... that precludes it being part of an imagestream, but does not necessarily mean they are not in the payload (though I would lean as well to your point that it means they are not).  But details to sort out in the 4.11 time frame, but I agree that assumption is the proper starting point.

> 
> > Next question, does that mean we still pursue things through ART, since shared resources is part of the payload, regardless of where our must-gather image lands, or is this a CPaaS item. Based on the recent 4.11 kick off, that has me leaning toward guessing CPaaS.
> 
> I think the long term strategy is to lessen the footprint of OpenShift - net new things should strive to be delivered on top with an operator. This is an extension to an item that exists in the payload and will be lifecycled with OpenShift, so I my first guess is that it will likewise go in the payload.

yep again the reasonable assumption to start things with, and then we can fully vet during the 4.11 timeframe

> 
> > That said, after this PR merges, I think we can still pursue openshift/release work for the new image, even before any payload/ART/CPaas decisions, like what occurred with shared resources. IIRC, that at least gets us a quay.io/openshift/origin-... image that we could direct customers too in a pinch.
> 
> Agree here - we'll want to build the new image in CI and mirror it to quay.io before doing anything else.

cool

Dockerfile.mustgather
gabemontero: this is the pattern exhibited elsewhere in multiple repos.

it *could* be an oversight that has been copy/pasted, but my assumption was there were elements in the must-gather image that are needed so that `oc adm must-gather` seemlessly calls this when used in conjunction with other must gather image.  You can call `oc adm must-gather` where you specify both the base must-gather imagestream, plus a feature image, as well as calling it with just the feature image.

if we want to fully satisfy your curiosity here, sure, and can do some experiments where I don't use multi-stage builds, and see what works / what does not work.

gabemontero: Before I try several permutations @adambkaplan do you have a precise Dockerfile you think should work?

gabemontero: So I tried 4 permutations.  2 failed on the image build.  Two at least built as images, but failed on execution.

```
FROM registry.ci.openshift.org/ocp/4.10:cli
WORKDIR /go/src/github.com/openshift/csi-driver-shared-resource
COPY . .
RUN chmod +x /go/src/github.com/openshift/csi-driver-shared-resource/must-gather/gather

ENTRYPOINT /go/src/github.com/openshift/csi-driver-shared-resource/must-gather/gather
```
Resulted in this error message:

`[must-gather-c9hxz] POD 2022-01-25T16:08:10.164298602Z /bin/bash: /usr/bin/gather: No such file or directory`

```
FROM registry.ci.openshift.org/ocp/4.10:must-gather
WORKDIR /go/src/github.com/openshift/csi-driver-shared-resource
COPY . .
RUN chmod +x /go/src/github.com/openshift/csi-driver-shared-resource/must-gather/gather

ENTRYPOINT /go/src/github.com/openshift/csi-driver-shared-resource/must-gather/gather

```

did not produced any explicit error messages I could find, but the data in question was never collected.

Unless you have a Dockerfile different than one of those you think should work @adambkaplan that I can try, I believe we can claim due diligence here and accept the multi-stage build pattern I picked up from the other examples.

thanks

adambkaplan (approved):
/lgtm

gabemontero:
/label px-approved

will help support

gabemontero:
/label qe-approved

until I've updated openshift/release there is nothing quasi consumable for QE

PR #87, opened 2022-01-28
Simplify the must-gather dockerfile

bparees:


bparees:
@gabemontero i believe this is all you need, though i admit i may be forgetting some nuance of the CI image build environment.

gabemontero:
> @gabemontero i believe this is all you need, though i admit i may be forgetting some nuance of the CI image build environment.

I believe I tried that permutation but maybe there was a slight diff in my attempt (which I no longer have) and this one.  Will report status of end to end testing shortly @bparees

Dockerfile.mustgather
```diff
@@ -1,9 +1,5 @@
-FROM registry.ci.openshift.org/ocp/4.10:must-gather AS builder
-WORKDIR /go/src/github.com/openshift/csi-driver-shared-resource
-COPY . .
-
-FROM registry.ci.openshift.org/ocp/4.10:cli
-COPY --from=builder /go/src/github.com/openshift/csi-driver-shared-resource/must-gather/* /usr/bin/
+FROM registry.ci.openshift.org/ocp/4.10:must-gather
+COPY must-gather/* /usr/bin/
```
gabemontero, line 2: I think I did this incorrectly when I tried this single stage permutation

gabemontero:
Yep it works @bparees thanks !

@adambkaplan FYI

gabemontero:
expedited the no-FF labeling given the prior PR which @bparees has simplified the Dockerfile for

gabemontero:
ah soot ... unrelated e2e flake ... suspect we won't make code freeze for 4.10 now @bparees 

think that means I'll need to adjust the openshift/release PR to not update the 4.10 branch

gabemontero:
confirmed this did not get forwarded into 4.10

it also as of this moment has not been forwarded into 4.11 yet

PR #93, opened 2022-04-14
use k8s atomic writer for storage of secrets / configmaps 

gabemontero:
Taking a glance at the latest shared resource epics, I cannot find a story related to this (@coreydaley @adambkaplan do let me know if that is just an oversight on my part), but an item we have discussed for completing productization of shared resource is switching actual storage of the secret/configmap keys to using [k8s atomic writer](https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/util/atomic_writer.go) to provide "atomic" access to data via some creative linux symlinking  so that if the contents change, users can use `inotify` or `fanotify` to coordinate access.

This is how a Pod's serviceaccount secrets have been stored since the early days of k8s, and this approach was created after some early bugs where intermittent errors would occur when read and writes of that secret data were done at the same time.

The upstream Secrets Store CSI Driver also adopted this approach somewhat recently.

As I mentioned last time we discussed this in team meetings, they actually create a copy of atomic_writer.go in their tree (even though they still already vendor in k8s/k8s).  In looking at the history of https://github.com/kubernetes-sigs/secrets-store-csi-driver/commits/main/pkg/util/fileutil/atomic_writer.go their updates from their initial copy are only superficial.

So I eschewed their copy approach and simply vendor in k8s/k8s like we do for openshift/builder and use the k8s/k8s atomic_writier.

Lastly, I moved our simple config map example to a locally created config map vs. one in the ocp config managed namespace.  It made it eaiser for manual testing around removing / adding keys and confirming that atomic_writer "handles it".

/assign @coreydaley 
/assign @adambkaplan


pkg/csidriver/driver.go
```diff
@@ -22,18 +22,19 @@ import (
...
 	"strings"
 	"sync"
+	"syscall"
 
 	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
+	atomic "k8s.io/kubernetes/pkg/volume/util"
```
coreydaley, line 36: I think that this should be something like "volumeutil", "kvolumeutil" or just "kutil" to be more descriptive to the package it is pulling in, not what feature of that package you are using.

pkg/csidriver/driver.go
```diff
@@ -244,38 +245,43 @@ func commonUpsertRanger(dv *driverVolume, key, value interface{}) error {
...
 	// around host filesystem issues might already exist would be warranted with such an exploration/effort
 
-	// Next, on an update we first nuke any existing directory and then recreate it to simplify handling the case where
-	// the keys in the secret/configmap have changed such that some keys have been removed, which would translate
-	// in files having to be removed. commonOSRemove will handle shares mounted off of shares.  And a reminder,
-	// currently this driver does not support overlaying over directories with files.  Either the directory in the
-	// container image must be empty, or the directory does not exist, and is created for the Pod's container as
-	// part of provisioning the container.
-	if err := commonOSRemove(podPath, fmt.Sprintf("commonUpsertRanger key %s volid %s share id %s pod name %s", key, dv.GetVolID(), dv.GetSharedDataId(), dv.GetPodName())); err != nil {
+	// NOTE: atomic_writer handles any pruning of secret/configmap keys where were present before, but are no longer
```
coreydaley, line 248: s/where/that/

pkg/csidriver/driver.go
```diff
@@ -244,38 +245,43 @@ func commonUpsertRanger(dv *driverVolume, key, value interface{}) error {
...
+				Data: content,
+				Mode: 0644,
 			}
 		}
 	}
+	if len(podFile) > 0 {
+		err = aw.Write(podFile)
+		if err != nil {
+			return err
+		}
```
coreydaley, line 283:
```
if err = aw.Write(podFile); err != nil {
	return err
}
```

pkg/csidriver/driver.go
```diff
@@ -285,20 +291,48 @@ func commonOSRemove(dir, dbg string) error {
...
 	// off of the potential mount point
-	return filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
-		if info == nil {
-			return nil
-		}
-		// since we do not support mounting on existing content, a dir can only mean a share
-		// has been mounted as a separate dir in our share, so skip
-		if info.IsDir() {
-			return nil
+	dirFile, err := os.Open(dir)
```
coreydaley, line 294: Should there be a deferred close directory after this?

pkg/csidriver/driver.go
```diff
@@ -285,20 +291,48 @@ func commonOSRemove(dir, dbg string) error {
...
+		}
+		// then we delete any non symlinks
+		if dirEntry.Type()&os.ModeSymlink == 0 {
+			fileName := filepath.Join(dir, dirEntry.Name())
+			klog.V(4).Infof("commonOSRemove going to delete file %s", fileName)
+			err = os.RemoveAll(fileName)
+			if err != nil {
+				klog.V(0).Infof("commonOSRemove: %s", err.Error())
+				return err
+			}
```
coreydaley, line 332:
```
if err = os.RemoveAll(fileName); err != nil {
klog.V(0).Infof("commonOSRemove: %s", err.Error())
return err
}
```
If this is going to be loglevel 0, it should probably be more human-friendly text, or maybe you meant it to be loglevel 4 like the others?

pkg/csidriver/driver_test.go
```diff
@@ -825,14 +825,21 @@ func findSharedItems(t *testing.T, dir string) (bool, bool) {
...
 			return nil
 		}
-		t.Logf("found file %s dir flag %v", info.Name(), info.IsDir())
-		if err == nil && strings.Contains(info.Name(), secretkey1) {
+		searchName := info.Name()
+		if info.Mode()&os.ModeSymlink != 0 {
+			linkName, err := os.Readlink(info.Name())
+			if err == nil {
+				searchName = linkName
+			}
```
coreydaley, line 833:
```
if linkName, err := os.Readlink(info.Name()); err == nil {
    searchName = linkName
}
```

pkg/csidriver/driver.go
```diff
@@ -244,38 +245,43 @@ func commonUpsertRanger(dv *driverVolume, key, value interface{}) error {
...
+	if err := os.MkdirAll(podPath, os.ModePerm); err != nil {
 		return err
 	}
-	if err := os.MkdirAll(podPath, os.ModePerm); err != nil {
+	podFile := map[string]atomic.FileProjection{}
+	aw, err := atomic.NewAtomicWriter(podPath, "shared-resource-csi-driver")
+	if err != nil {
 		return err
 	}
 	if payload.ByteData != nil {
```
coreydaley, line 258: Can a `payload` have both `ByteData` and `StringData` populated at the same time? Can `payload` have both be `nil` also?

pkg/csidriver/driver.go
```diff
@@ -285,20 +291,48 @@ func commonOSRemove(dir, dbg string) error {
...
+	}
+	// we have to unlink the key from our configmap/secret first with the atomic writer symlinks
+	for _, dirEntry := range dirEntries {
+		if dirEntry.Type()&os.ModeSymlink != 0 && !strings.HasPrefix(dirEntry.Name(), "..") {
+			// unlink our secret/configmap key
+			err = syscall.Unlink(filepath.Join(dir, dirEntry.Name()))
+			if err != nil {
+				klog.V(0).Infof("commonOSRemove: %s", err.Error())
+				return err
+			}
```
coreydaley, line 310:
```
if err = syscall.Unlink(filepath.Join(dir, dirEntry.Name())); err != nil {
    klog.V(0).Infof("commonOSRemove: %s", err.Error())
    return err
}
```

This should probably be loglevel 4 or more human-friendly text.

pkg/csidriver/driver.go
```diff
@@ -285,20 +291,48 @@ func commonOSRemove(dir, dbg string) error {
...
-	})
+	}
+	// we then unlink the symlink from atomic writer that start with ".."
+	for _, dirEntry := range dirEntries {
+		if dirEntry.Type()&os.ModeSymlink != 0 && strings.HasPrefix(dirEntry.Name(), "..") {
+			err = syscall.Unlink(filepath.Join(dir, dirEntry.Name()))
+			if err != nil {
+				klog.V(0).Infof("commonOSRemove: %s", err.Error())
+				return err
+			}
```
coreydaley, line 321:
```
if err = syscall.Unlink(filepath.Join(dir, dirEntry.Name())); err != nil {
    klog.V(0).Infof("commonOSRemove: %s", err.Error())
    return err
}
```
Loglevel 4 or more human-friendly text

coreydaley (changes requested):

gabemontero:
Compile errors I am not seeing locally:

```
Running e2e suite normal
# sigs.k8s.io/json/internal/golang/encoding/json
vendor/sigs.k8s.io/json/internal/golang/encoding/json/encode.go:1249:12: sf.IsExported undefined (type reflect.StructField has no field or method IsExported)
vendor/sigs.k8s.io/json/internal/golang/encoding/json/encode.go:1255:18: sf.IsExported undefined (type reflect.StructField has no field or method IsExported)
FAIL	github.com/openshift/csi-driver-shared-resource/test/e2e [build failed]
FAIL
```

probably need to bump the golang version in the Dockerfile and/or openshift/release

gabemontero:
weirdly the unit test at least is using 1.16 which definitely has the issue:

[36mINFO[0m[2022-04-14T16:22:34Z] Tagging openshift/release:golang-1.16 into pipeline:root. 

e2e's are 1.17 ... though I cannot tell with dot rel off of 1.17

[36mINFO[0m[2022-04-14T16:22:35Z] Tagging ocp/builder:rhel-8-golang-1.17-openshift-4.11 into pipeline:ocp_builder_rhel-8-golang-1.17-openshift-4.11.

gabemontero:
always forget the ci-operator yam file ... that at least fixed the unit test

gabemontero:
removed permissions needs a little work ; my simple test did not catch what the more complicated e2e caught

will pick back up after the holiday

golang version at least sorted out

pkg/csidriver/driver.go
gabemontero: yes and yes

see the ConfigMap godoc wrt both populated 

both nil means an empty secret/configmap

gabemontero:
hmm ... slow passes for me locally

/test e2e-aws-csi-driver-slow

gabemontero:
> hmm ... slow passes for me locally
> 
> /test e2e-aws-csi-driver-slow

reviewed ... the test waits 10 min and the relist is 10 min ... could have hit edge case.  May push tweak to wait time based on next set of results.

gabemontero:
ok e2e's all green

doing one more run with the slow e2e interval tweaked

@coreydaley have the updates from your comments in a separate commit

PTAL / thanks

test/e2e/slow_test.go
```diff
@@ -20,14 +20,14 @@ func TestBasicThenNoRBACThenRBAC(t *testing.T) {
...
+	t.Logf("%s: wait up to 11 minutes for examining pod %s since the controller does not currently watch all clusterroles and clusterrolebindings and reverse engineer which ones satisfied the SAR calls, so we wait for relist on shares", time.Now().String(), testArgs.Name)
 	testArgs.SearchStringMissing = true
-	testArgs.TestDuration = 10 * time.Minute
+	testArgs.TestDuration = 11 * time.Minute
 	testArgs.SearchString = "invoker"
 	framework.ExecPod(testArgs)
 
 	framework.CreateShareRelatedRBAC(testArgs)
-	t.Logf("%s: wait up to 10 minutes for examining pod %s since the controller does not currently watch all clusterroles and clusterrolebindings and reverse engineer which ones satisfied the SAR calls, so we wait for relist on shares", time.Now().String(), testArgs.Name)
+	t.Logf("%s: wait up to 11 minutes for examining pod %s since the controller does not currently watch all clusterroles and clusterrolebindings and reverse engineer which ones satisfied the SAR calls, so we wait for relist on shares", time.Now().String(), testArgs.Name)
```
coreydaley, line 30: Does this need the `testArgs.TestDuration` or will it roll over from the previous one?

gabemontero: rolls over from previous

coreydaley: Ok, thanks.

gabemontero:
> /lgtm

thanks @coreydaley 

that said,

/hold

have to squash your review commit

gabemontero:
/label docs-approved

no docs impact - internal implementation detail - also see relative lack on docs in upstream k8s around using inotify, etc. ... considered a "linux" thing

gabemontero:
/label px-approved

internal implementation detail

gabemontero:
launching clusterbot cluster with this PR for final verification of entitled build usage

gabemontero:
forgot, cannot do entitled builds with clusterbot clusters

had to verify manually via local build ... will do so again after this merges and we have a level with it

/label qe-approved

gabemontero:
verified openshift builds with entitlements at level 4.11.0-0.ci-2022-04-18-211945

PR #94, opened 2022-04-18
update test pod to alleviate PodSecurity warnings

gabemontero:
during some recent testing I noticed this message in the e2e's:

```
W0415 16:29:54.714771   12309 warnings.go:70] would violate PodSecurity "restricted:latest": allowPrivilegeEscalation != false (container "my-frontend" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (container "my-frontend" must set securityContext.capabilities.drop=["ALL"]), runAsNonRoot != true (pod or container "my-frontend" must set securityContext.runAsNonRoot=true), seccompProfile (pod or container "my-frontend" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
```

So I updated the pod spec to address.

Then, I had to move off of `registry.redhat.io/ubi8/ubi` because it ran as root.

Finally, our test image did not support `ls -laRZ` so I dropped `Z`.  Now that we only support read-only volumes and SELinux is thus sorted out, `Z` is not longer necessary.

/assign @coreydaley

### DIFF
--- a/test/framework/pod.go
+++ b/test/framework/pod.go
@@ -29,6 +29,7 @@ const (
 func CreateTestPod(t *TestArgs) {
 	t.T.Logf("%s: start create test pod %s", time.Now().String(), t.Name)
 	truVal := true
+	falVal := false
 	pod := &corev1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      t.Name,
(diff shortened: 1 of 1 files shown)


gabemontero:
/label px-approved

no impact

gabemontero:
/label qe-approved

no impact

gabemontero:
/label docs-approved

no impact

PR #96, opened 2022-04-29
add Tekton example, update/correction to README and existing example docs

gabemontero:
The tekton example was one I had on my queue, based on some early pipeline service discussions I had, and then was accelerated by some conference calls today, as well as https://coreos.slack.com/archives/C02FTKEMBU1/p1651237549248689 from @sbose78 

@vdemeester @Michkov FYI

per the slack thread above, once the particulars needed for this work are understood here, the next step per @sbose78 is replicating this somewhere under https://github.com/redhat-appstudio/build-definitions/tree/main/pipelines/base

the use case first was cited by @sbose78 in https://github.com/redhat-appstudio/application-service/pull/65

### DIFF
--- a/examples/tekton/01-clusterrole-privileged.yaml
+++ b/examples/tekton/01-clusterrole-privileged.yaml
@@ -1,0 +1,11 @@
+# When https://issues.redhat.com/browse/STOR-828 is implemented in OpenShift, and the Shared Resource CSI Driver can be
+# deemed safe for restricted users, this ClusterRole and its associated binding can be removed.
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: my-shared-resource-privileged-role
+rules:
+  - apiGroups: ["security.openshift.io"]
+    resourceNames: ["my-csi-scc-from-restricted"]
+    resources: ["securitycontextconstraints"]
+    verbs: ["use"]
--- a/examples/tekton/01-rolebinding-share.yaml
+++ b/examples/tekton/01-rolebinding-share.yaml
@@ -1,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: shared-resource-my-share-tekton-default
+  namespace: my-csi-app-namespace
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: shared-resource-my-share-tekton
+subjects:
+  - kind: ServiceAccount
+    name: default
+    namespace: my-csi-app-namespace
--- a/README.md
+++ b/README.md
@@ -34,8 +34,8 @@ NOTE: see [CSI Volume Specifics](docs/csi.md) for restrictions around these feat
 ## Getting Started
 
 The maintenance of the related API objects and the deployment of this CSI driver are handled via the [Openshift CSI Driver for Shared Resources Operator](https://github.com/openshift/csi-driver-shared-resource-operator)
-when you are using a Tech Preview OpenShift Cluster in 4.10.  The 4.10 release docs are not out yet with 4.10 still under development,
-but these [4.9 docs](https://docs.openshift.com/container-platform/4.9/post_installation_configuration/cluster-tasks.html#post-install-tp-tasks) are 
+when you are using a Tech Preview OpenShift Cluster in 4.10.  The 4.10 release docs are [here](https://docs.openshift.com/container-platform/4.10/storage/container_storage_interface/ephemeral-storage-shared-resource-csi-driver-operator.html),
+and these [4.10 docs](https://docs.openshift.com/container-platform/4.10/post_installation_configuration/cluster-tasks.html#post-install-tp-tasks) are 
 sufficient for explaining how to turn on Tech Preview features after install.
 
 For running on a 4.10 cluster which is *NOT* a Tech Preview cluster, you must employ the methodology described in the [Local Devlopment](#local-development)
--- a/docs/build-with-rhel-entitlements.md
+++ b/docs/build-with-rhel-entitlements.md
@@ -1,6 +1,14 @@
 # BuildConfig example
 
-From the root directory, create/apply the various API objects defined in the YAML files in the `./examples` directory.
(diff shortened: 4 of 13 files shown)


examples/tekton/01-rolebinding-share.yaml
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  name: shared-resource-my-share-tekton-default
+  namespace: my-csi-app-namespace
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: shared-resource-my-share-tekton
+subjects:
+  - kind: ServiceAccount
+    name: default
```
sbose78, line 12: Is this service account that the TaskRun pod runs as ?

gabemontero: yes it is

and I made a conscious decision to use `default` instead of `pipeline`:

- although shared resources is still a "downstream" thing, whenever possible I try to make it as "upstream friendly" as I can
- I also had vague recollections about generally wanting to move away from the `pipeline` SA (but of course I may have incorrect or out of date recollections here)
- and of course, switching the `pipeline` here, or simply using `pipeline` in whatever version of this might land in app studio, should be straight forward

examples/tekton/01-clusterrole-privileged.yaml
```diff
@@ -0,0 +1,11 @@
+# When https://issues.redhat.com/browse/STOR-828 is implemented in OpenShift, and the Shared Resource CSI Driver can be
+# deemed safe for restricted users, this ClusterRole and its associated binding can be removed.
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: my-shared-resource-privileged-role
+rules:
+  - apiGroups: ["security.openshift.io"]
+    resourceNames: ["privileged"]
```
gabemontero, line 9: yes, creating a new SCC that adds the `csi` volume type but otherwise minimizes where ever possible the power of the SCC we must define before STOR-828 lands narrows the exposure somewhat, but that said, exposing CSI drivers at all leaves the exposure big enough that it is more or less a drop in the bucket in my opinion

but we can go down the seaprate SCC path in app studio for sure @sbose78 .... again, I tried to keep things a bit simple here

gabemontero: that said, I just try a "restricted scc + csi volumes" role and it does work

I'll update the PR in a bit

gabemontero:
terraform flake on e2e-aws-csi-driver

/retest

gabemontero:
/label px-approved

no px impact - example only

gabemontero:
/label docs-approved

@rolfedh FYI - the openshift pipelines team has https://issues.redhat.com/browse/SRVKP-1025 on their plate for documenting the use of shared resources in tekton.  That epic cites this PR as a reference, but given how examples like this are "pipeline centric" and part of their catalog, where they make updates to pipeline core when needed to "make things easier", let's wait until they take things on, in case it changes how this should be documented

so for us, it is for now a no-op.

gabemontero:
I'm going to try and have a quick chat with @prietyc123 and Shwetha on this before self applying the qe approved label for this sample, though one could argue it falls under the domain of https://issues.redhat.com/browse/SRVKP-1025 as well.

vdemeester (commented):
@gabemontero we could also have an example of using this with `workspace` as well, right ?
(if workspace doesn't allow `csi` to be used, this is definitely a feature we would need to work on, I'll take a look 👼🏼 )

examples/tekton/01-clusterrole-privileged.yaml
sbose78: ah, great 😎

examples/tekton/01-rolebinding-share.yaml
sbose78: Sounds good, Gabe!

gabemontero:
> @gabemontero we could also have an example of using this with `workspace` as well, right ? (if workspace doesn't allow `csi` to be used, this is definitely a feature we would need to work on, I'll take a look 👼🏼 )

fyi @vdemeester is taking on the upstream tekton issue that will add csi for workspaces in tekton ... mid summer estimate for it landing in openshift pipelines (which will be 1.37 or 1.38 tekton ... the feature is targeted for 1.36)

that said, we'll merge this in the interim and update later when the support arrives ... also, @sbose78 would like me to pursue the non-workspace approach in app studio in the interim

gabemontero:
given what @prietyc123 has going on before she goes on leave, and that we'll revisit this sample again once the tekton workspace support is there, so she can tackle smoke tests then, I'm going ahead and 

/label qe-approved

just to facilitate merge for reference by the pipieline service team ... again, this is not something that is supported / documented / integrated as part of 4.11.  Purely an upstream developer example to illustrate current status.

gabemontero:
/assign @coreydaley 

sufficient feedback received from stakeholders @sbose78 and @vdemeester 

just need a generic once over for clean up items, then we can merge.  See prior comments on the dev only nature of this.

thanks

rolfedh:
> @rolfedh FYI - the openshift pipelines team has https://issues.redhat.com/browse/SRVKP-1025 on their plate for documenting the use of shared resources in tekton. That epic cites this PR as a reference, but given how examples like this are "pipeline centric" and part of their catalog, where they make updates to pipeline core when needed to "make things easier", let's wait until they take things on, in case it changes how this should be documented
> 
> so for us, it is for now a no-op.

@Preeticp @sounix000 FYI

gabemontero:
On Tue, May 3, 2022 at 12:45 AM Vincent Demeester ***@***.***>
wrote:

> ***@***.**** commented on this pull request.
>
> @gabemontero <https://github.com/gabemontero> we could also have an
> example of using this with workspace as well, right ?
> (if workspace doesn't allow csi to be used, this is definitely a feature
> we would need to work on, I'll take a look 👼🏼 )
>

Yep I looked and workspace does NOT allow csi @vdemeester

The notion of addressing that upstream did occur to me and I had planned on
pinging you about it.  One potential concern is the pod security
implications.  I will ping you in slack in a bit to elaborate on that.

PR #109, opened 2022-07-12
[WIP] BUILD-407 Reserved name prefix for shared resources

prietyc123:


prietyc123:
@coreydaley I have submitted the implementation code and need to add the unit test to verify the behavior. Could you please have a look into the code, meanwhile I will add the tests.

prietyc123:
Aligned with the fix https://github.com/openshift/csi-driver-shared-resource-operator/pull/53

pkg/webhook/csidriver/csidriver.go
```diff
@@ -81,6 +101,72 @@ func (s *SharedResourcesCSIDriverWebhook) authorizePod(request admissionctl.Requ
...
+			if strings.Contains(value, SharedSecret.ObjectMeta.Name) {
+				allowResource = true
+				break
+			}
+		}
+		if !allowResource {
+			admissionctl.Denied("Shared secrets are not allowed to be created with the reserved prefix \"openshift-\" unless its name and Reference key are found in a well known list of OCP pre-populated Shared Resource")
+			ret.UID = request.AdmissionRequest.UID
+			return ret
+		}
```
coreydaley, line 128: You should be able to refactor this into an additional utility method that can be used here and in the `authorizedSharedConfigMap` method, that will cut down on the amount of code that needs to be maintained and ensure that it acts the same.

pkg/webhook/csidriver/csidriver.go
```diff
@@ -81,6 +101,72 @@ func (s *SharedResourcesCSIDriverWebhook) authorizePod(request admissionctl.Requ
...
+
+	if strings.HasPrefix(SharedSecret.ObjectMeta.Name, "openshift-") {
+		for _, value := range config.LoadedConfig.AllowedSharedResources {
+			if strings.Contains(value, SharedSecret.ObjectMeta.Name) {
+				allowResource = true
+				break
+			}
+		}
+		if !allowResource {
+			admissionctl.Denied("Shared secrets are not allowed to be created with the reserved prefix \"openshift-\" unless its name and Reference key are found in a well known list of OCP pre-populated Shared Resource")
```
coreydaley, line 125: This is probably a bit long and verbose, I think it can be cut down to a simplified message.

coreydaley (changes requested):

coreydaley:
@prietyc123 Looking pretty good so far, commented about a few refactors and such.

prietyc123:
/assign @coreydaley 
/assign @gabemontero 

For PR review. @coreydaley I have addressed the comments you posted 🙂

pkg/webhook/csidriver/csidriver.go
```diff
@@ -81,6 +101,59 @@ func (s *SharedResourcesCSIDriverWebhook) authorizePod(request admissionctl.Requ
...
+func (s *SharedResourcesCSIDriverWebhook) authorizesharedConfigMap(request admissionctl.Request) admissionctl.Response {
+	klog.V(2).Info("admitting shared configmap with SharedResourceCSIDiver")
+	SharedConfigMap, err := s.renderSharedConfigmap(request)
+	if err != nil {
+		klog.Error(err, "Couldn't render a shared configmap from the incoming request")
+		return admissionctl.Errored(http.StatusBadRequest, err)
+	}
+
+	ret := s.allowResource(request, SharedConfigMap.ObjectMeta.Name)
+	return ret
```
coreydaley, line 126:
```suggestion
	return s.allowResource(request, SharedConfigMap.ObjectMeta.Name)
```

pkg/webhook/csidriver/csidriver.go
```diff
@@ -81,6 +101,59 @@ func (s *SharedResourcesCSIDriverWebhook) authorizePod(request admissionctl.Requ
...
+}
+
+// allowResource checks whether the shared resources can be created
+// with "openshift-" prefix or not.
+func (s *SharedResourcesCSIDriverWebhook) allowResource(request admissionctl.Request, resource string) admissionctl.Response {
```
coreydaley, line 131: This would probably make more sense to be named `authorizeResource`

pkg/webhook/csidriver/csidriver.go
```diff
@@ -81,6 +101,59 @@ func (s *SharedResourcesCSIDriverWebhook) authorizePod(request admissionctl.Requ
...
+	if strings.HasPrefix(resource, "openshift-") {
+		for _, value := range config.LoadedConfig.AllowedSharedResources {
+			if strings.Contains(value, resource) {
+				allowResource = true
+				break
+			}
+		}
+		if !allowResource {
+			admissionctl.Denied("Shared resource with prefix 'openshift-' is not allowed, unless listed under OCP pre-populated shared resource")
+			return ret
```
coreydaley, line 147:
```suggestion
			ret = admissionctl.Denied("Shared resource with prefix 'openshift-' is not allowed, unless listed under OCP pre-populated shared resource")
			ret.UID = request.AdmissionRequest.UID
			return ret
```

pkg/webhook/csidriver/csidriver.go
```diff
@@ -98,3 +171,41 @@ func (s *SharedResourcesCSIDriverWebhook) renderPod(request admissionctl.Request
 
 	return pod, err
 }
+
+// renderSharedConfigmap decodes an *v1alpha1.SharedConfigMap from the incoming request.
+// If the request includes an OldObject (from an update or deletion), it will be
+// preferred, otherwise, the Object will be preferred.
+func (s *SharedResourcesCSIDriverWebhook) renderSharedConfigmap(request admissionctl.Request) (*v1alpha1.SharedConfigMap, error) {
```
coreydaley, line 178:
```suggestion
func (s *SharedResourcesCSIDriverWebhook) renderSharedConfigMap(request admissionctl.Request) (*v1alpha1.SharedConfigMap, error) {
```

pkg/webhook/csidriver/csidriver.go
```diff
@@ -81,6 +101,59 @@ func (s *SharedResourcesCSIDriverWebhook) authorizePod(request admissionctl.Requ
...
+		return admissionctl.Errored(http.StatusBadRequest, err)
+	}
+
+	ret := s.allowResource(request, SharedSecret.ObjectMeta.Name)
+	return ret
+}
+
+func (s *SharedResourcesCSIDriverWebhook) authorizesharedConfigMap(request admissionctl.Request) admissionctl.Response {
+	klog.V(2).Info("admitting shared configmap with SharedResourceCSIDiver")
+	SharedConfigMap, err := s.renderSharedConfigmap(request)
```
coreydaley, line 119:
```suggestion
	SharedConfigMap, err := s.renderSharedConfigMap(request)
```

pkg/webhook/csidriver/csidriver.go
```diff
@@ -98,3 +171,41 @@ func (s *SharedResourcesCSIDriverWebhook) renderPod(request admissionctl.Request
...
+// preferred, otherwise, the Object will be preferred.
+func (s *SharedResourcesCSIDriverWebhook) renderSharedConfigmap(request admissionctl.Request) (*v1alpha1.SharedConfigMap, error) {
+	decoder, err := admissionctl.NewDecoder(scheme)
+	if err != nil {
+		return nil, err
+	}
+
+	sharedConfigmap := &v1alpha1.SharedConfigMap{}
+	if len(request.OldObject.Raw) > 0 {
+		err = decoder.DecodeRaw(request.OldObject, sharedConfigmap)
```
coreydaley, line 186:
```suggestion
		err = decoder.DecodeRaw(request.OldObject, sharedConfigMap)
```

coreydaley (changes requested):

pkg/config/config.go
```diff
@@ -33,7 +36,8 @@ func (c *Config) GetShareRelistInterval() time.Duration {
 // NewConfig returns a Config instance using the default attribute values.
 func NewConfig() Config {
 	return Config{
-		ShareRelistInterval: DefaultResyncDuration.String(),
-		RefreshResources:    true,
+		AllowedSharedResources: []string{},
```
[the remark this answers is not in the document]
gabemontero: would this not be easier to use if it was `map[string]string` where the key is the shared secret or shared config map name, and the value is the namespace name : instance name combo?

gabemontero (changes requested):
I would expect and e2e test that attempts to create a shared secret named `openshift-etc-pki-entitlement`
that points the a namespace/name other than the one specfied at https://github.com/openshift/csi-driver-shared-resource-operator/pull/53/files#diff-22da0d537c16c5218fec4a6eb792c4d5c872d215595a01f7edecbf5176a1f033R36

I would also expect a check near https://github.com/openshift/csi-driver-shared-resource/blob/a9a21b7a1f9ca6c898b585d92c68d5655eefd149/pkg/csidriver/nodeserver.go#L197-L199 to cover the case that the webhook is down, per what is expected in the enhancement proposal.

i.e. of a shared resource starts with `openshift-` and is not in our config, don't allow it.

pkg/config/config.go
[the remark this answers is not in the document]
coreydaley: That would work as long as it will load correctly from the config file.

coreydaley:
> I would expect and e2e test that attempts to create a shared secret named `openshift-etc-pki-entitlement` that points the a namespace/name other than the one specfied at https://github.com/openshift/csi-driver-shared-resource-operator/pull/53/files#diff-22da0d537c16c5218fec4a6eb792c4d5c872d215595a01f7edecbf5176a1f033R36
> 
> I would also expect a check near
> 
> https://github.com/openshift/csi-driver-shared-resource/blob/a9a21b7a1f9ca6c898b585d92c68d5655eefd149/pkg/csidriver/nodeserver.go#L197-L199
> to cover the case that the webhook is down, per what is expected in the enhancement proposal.
> 
> i.e. of a shared resource starts with `openshift-` and is not in our config, don't allow it.

I think @prietyc123 is trying to iron out the implementation first so that she can design the proper tests, this is really just the first iteration of the pull request.

gabemontero:
> > I would expect and e2e test that attempts to create a shared secret named `openshift-etc-pki-entitlement` that points the a namespace/name other than the one specfied at https://github.com/openshift/csi-driver-shared-resource-operator/pull/53/files#diff-22da0d537c16c5218fec4a6eb792c4d5c872d215595a01f7edecbf5176a1f033R36
> > I would also expect a check near
> > https://github.com/openshift/csi-driver-shared-resource/blob/a9a21b7a1f9ca6c898b585d92c68d5655eefd149/pkg/csidriver/nodeserver.go#L197-L199
> > 
> > to cover the case that the webhook is down, per what is expected in the enhancement proposal.
> > i.e. of a shared resource starts with `openshift-` and is not in our config, don't allow it.
> 
> I think @prietyc123 is trying to iron out the implementation first so that she can design the proper tests, this is really just the first iteration of the pull request.

OK good to know thanks @coreydaley 

By proper tests I assume you are referring to the smoke tests.  Which is good for sure, but yeah ultimately we want both e2e and smoke tests.

gabemontero:
and yeah just wanted to make sure we don't skip on the check in the driver itself even though we have the webhook now.

prietyc123:
> and yeah just wanted to make sure we don't skip on the check in the driver itself even though we have the webhook now.

@gabemontero Could you please elaborate more on what do you mean by adding checks in the driver. Am I missing something here.

prietyc123:
> > and yeah just wanted to make sure we don't skip on the check in the driver itself even though we have the webhook now.
> 
> @gabemontero Could you please elaborate more on what do you mean by adding checks in the driver. Am I missing something here.

Ahh nevermind I got it. I skipped your comment in https://github.com/openshift/csi-driver-shared-resource/pull/109#pullrequestreview-1045581455 thats why was not getting the clear picture.

pkg/webhook/csidriver/csidriver.go
```diff
@@ -81,6 +101,58 @@ func (s *SharedResourcesCSIDriverWebhook) authorizePod(request admissionctl.Requ
...
+			if strings.EqualFold(key, resource) {
+				allowResource = true
+				break
+			}
+		}
+		if !allowResource {
+			ret = admissionctl.Denied("Shared resource with prefix 'openshift-' is not allowed, unless listed under OCP pre-populated shared resource")
+			ret.UID = request.AdmissionRequest.UID
+			return ret
+		}
```
coreydaley, line 147:
```suggestion
		if _, ok := config.LoadedConfig.AuthorizedSharedResources[resource]; !ok {
			ret = admissionctl.Denied("Shared resource with prefix 'openshift-' is not allowed, unless listed under OCP pre-populated shared resource")
			ret.UID = request.AdmissionRequest.UID
			return ret
		}
```

coreydaley: You only need to check if the resource exists in the map, you don't need to iterate over the map like you would for a slice.

pkg/webhook/csidriver/csidriver.go
```diff
@@ -81,6 +101,49 @@ func (s *SharedResourcesCSIDriverWebhook) authorizePod(request admissionctl.Requ
...
+	}
+
+	// Hereafter, all requests are controlled
+	ret = admissionctl.Allowed("Allowed to create shared configmap")
+	ret.UID = request.AdmissionRequest.UID
+	return ret
```
coreydaley, line 144:
```suggestion
	ret.UID = request.AdmissionRequest.UID
	if strings.HasPrefix(resource, "openshift-") {
		if _, exists := config.LoadedConfig.AuthorizedSharedResources[resource]; !exists {
			ret = admissionctl.Denied("Shared resource with prefix 'openshift-' is not allowed, unless listed under OCP pre-populated shared resource")

			return ret
		}
	}

	// Hereafter, all requests are controlled
	ret = admissionctl.Allowed("Allowed to create shared configmap")
	return ret
```

pkg/csidriver/nodeserver.go
```diff
@@ -193,6 +194,15 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		return nil, err
 	}
 
+	s := csidriver.NewWebhook()
```
coreydaley, line 197: We should not create a new webhook here, You should move the `ListedShareResource` function to another package (maybe a util package or something) and share it between this file and the one for the csidriver webhook.

prietyc123: Make sense 👍

pkg/util/util.go
```diff
@@ -0,0 +1,18 @@
+package util
+
+import (
+	"strings"
+
+	"github.com/openshift/csi-driver-shared-resource/pkg/config"
+)
+
+// ListedShareResource verify the shared resource already exists under OCP pre-populated shared resource
+func ListedShareResource(resource, resourceRefKeyNm, resourceRefKeyNs string) bool {
```
coreydaley, line 10:
```suggestion
// IsAuthorizedSharedResource verify the shared resource already exists under OCP pre-populated shared resource
func IsAuthorizedSharedResource(resource, resourceRefKeyNm, resourceRefKeyNs string) bool {
```

coreydaley:
I think we need to sync back up on this next week. There should be two new configmaps created, one for the secrets, and one for the configmaps as per https://github.com/openshift/enhancements/blob/master/enhancements/cluster-scope-secret-volumes/shared-resource-validation.md#api-extensions

prietyc123:
> I think we need to sync back up on this next week. There should be two new configmaps created, one for the secrets, and one for the configmaps as per https://github.com/openshift/enhancements/blob/master/enhancements/cluster-scope-secret-volumes/shared-resource-validation.md#api-extensions

Yes I think you are right. I was also wondering that. Lets connect this coming monday.

coreydaley:
/retitle [WIP] BUILD-407: Reserved name prefix for shared resources

pkg/util/util.go
```diff
@@ -0,0 +1,26 @@
...
+	var (
+		value  string
+		exists bool
+	)
+	if strings.HasPrefix(resource, "openshift-") {
+		if strings.EqualFold(kind, "SharedSecret") {
+			value, exists = config.LoadedSharedSecretConfig.AuthorizedSharedResources[resource]
+		} else {
+			value, exists = config.LoadedSharedConfig.AuthorizedSharedResources[resource]
+		}
```
coreydaley, line 20: Let's use a switch/case statement here to facilitate adding more types in the future if needed

pkg/config/config.go
```diff
@@ -17,7 +17,25 @@ type Config struct {
 	RefreshResources bool `yaml:"refreshResources,omitempty"`
 }
 
-var LoadedConfig Config
+// SharedConfigList configuration attributes
+type SharedConfigList struct {
```
coreydaley, line 21:
```suggestion
type SharedConfigMaps struct {
```

pkg/config/config.go
```diff
@@ -17,7 +17,25 @@ type Config struct {
...
+	// AuthorizedSharedResources has the list of shared configmaps
+	// available in the list of OCP pre-populated Shared Resource
+	AuthorizedSharedResources map[string]string `yaml:"authorizedSharedResources,omitempty"`
+}
+
+// SharedSecretConfigList configuration attributes
+type SharedSecretConfigList struct {
```
coreydaley, line 28:
```suggestion
type SharedSecrets struct {
```

pkg/util/util.go
```diff
@@ -0,0 +1,27 @@
...
+		exists bool
+	)
+	if strings.HasPrefix(resource, "openshift-") {
+		switch kind {
+		case "SharedSecret":
+			value, exists = config.LoadedSharedSecrets.AuthorizedSharedResources[resource]
+		case "SharedConfigMap":
+			value, exists = config.LoadedSharedConfigMaps.AuthorizedSharedResources[resource]
+		}
+		if !exists || (exists && value != (resourceRefKeyNs+":"+resourceRefKeyNm)) {
```
coreydaley, line 22:
```suggestion
		if !exists || value != (resourceRefKeyNs+":"+resourceRefKeyNm) {
```
If the first part of the expression is false `!exists` it won't even check the second part, so there is no need for the `exists` check.

pkg/webhook/csidriver/csidriver.go
```diff
@@ -47,12 +49,29 @@ func (s *SharedResourcesCSIDriverWebhook) Name() string { return WebhookName }
...
-	return req.Kind.Kind == "Pod"
+	switch req.Kind.Kind {
+	case "Pod":
+		return req.Kind.Kind == "Pod"
+	case "SharedConfigMap":
+		return req.Kind.Kind == "SharedConfigMap"
+	case "SharedSecret":
+		return req.Kind.Kind == "SharedSecret"
+	}
+	return false
```
coreydaley, line 60:
```suggestion
	switch req.Kind.Kind {
	case "Pod":
		return req.Kind.Kind == "Pod"
	case "SharedConfigMap":
		return req.Kind.Kind == "SharedConfigMap"
	case "SharedSecret":
		return req.Kind.Kind == "SharedSecret"
	default:
	    return false
	}
```

pkg/webhook/csidriver/csidriver.go
```diff
@@ -47,12 +49,29 @@ func (s *SharedResourcesCSIDriverWebhook) Name() string { return WebhookName }
...
 }
 
 // Authorized implements Webhook interface
 func (s *SharedResourcesCSIDriverWebhook) Authorized(request admissionctl.Request) admissionctl.Response {
-	return s.authorizePod(request)
+	var ret admissionctl.Response
+	switch request.Kind.Kind {
```
coreydaley, line 66: You should probably add a default case here which returns a Denied response and logs some kind of warning about a Kind that is not handled

coreydaley (changes requested):

pkg/config/config.go
```diff
@@ -37,3 +55,12 @@ func NewConfig() Config {
...
 }
+
+// NewSharedSecrets returns a SharedSecret Configmap instance using the default attributes values.
+func NewSharedSecrets() SharedSecrets {
+	return SharedSecrets{
+		AuthorizedSharedResources: map[string]string{
+			"openshift-etc-pki-entitlement": "openshift-config-managed:etc-pki-entitlement",
+		},
+	}
+}
```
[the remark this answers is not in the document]
prietyc123: yes, you are right. I forgot to remove this.

pkg/webhook/csidriver/csidriver.go
```diff
@@ -47,12 +49,30 @@ func (s *SharedResourcesCSIDriverWebhook) Name() string { return WebhookName }
...
-	return s.authorizePod(request)
+	switch request.Kind.Kind {
+	case "Pod":
+		return s.authorizePod(request)
+	case "SharedConfigMap":
+		return s.authorizeSharedConfigMap(request)
+	case "SharedSecret":
+		return s.authorizeSharedSecret(request)
+	default:
+		return admissionctl.Denied("Requesting resource kind is unknown")
```
coreydaley, line 74:
```suggestion
		return admissionctl.Denied(fmt.Sprintf("Requesting resource kind %q is not supported", request.Kind.Kind))
```

prietyc123:
/retitle BUILD-407: Reserved name prefix for shared resources

prietyc123:
@coreydaley This pr is also ready for review. I have also ran the unit tests, I added in this pr with the change introduced via [operator repo pr](BUILD-407: Reserved name prefix for shared resources). All the tests run successfully. Also the files are reflected inside the cluster as per [comment](https://github.com/openshift/csi-driver-shared-resource-operator/pull/56#issuecomment-1274178054)

prietyc123:
Unit test scenario will fail untill we merge https://github.com/openshift/csi-driver-shared-resource-operator/pull/56

prietyc123:
bump @coreydaley

cmd/csidriver/main.go
```diff
@@ -130,6 +151,8 @@ func init() {
 	rootCmd.Flags().AddGoFlagSet(flag.CommandLine)
 
 	rootCmd.Flags().StringVar(&cfgFilePath, "config", "/var/run/configmaps/config/config.yaml", "configuration file path")
+	rootCmd.Flags().StringVar(&shCfgFilePath, "shConfig", "/var/run/configmaps/config/sharedconfigmap-list.yaml", "Shared config configuration file path")
+	rootCmd.Flags().StringVar(&shSecretCfgFilePath, "shSecret", "/var/run/configmaps/config/sharedsecret-list.yaml", "Shared secret configuration file path")
```
coreydaley, line 155: These need to be removed, and the `config` line above them needs to be updated to be a sliceVar.

prietyc123: Done

gabemontero:
Implemeted via https://github.com/openshift/csi-driver-shared-resource/pull/122

PR #127, opened 2023-02-14
Adding test scenariuos for Reserved Name Prefix for SharedSecret/SharedConfigMaps

jitendar-singh:
- Defined feature file with scenarios covering Reserved Name Prefix for SharedSecret and SharedConfigMaps
- Defined bad-shareconfig-name.yaml 
- Defined bad-sharesecrets-name.yaml


gabemontero (changes requested):
please add 1 more case @jitendar-singh 

add a shared secret with name `openshift-etc-pki-entitlement` and the secret it shares is the `etc-pki-entitlement` secret from the `openshift-config-managed` namespace

that is the one item we current register in the valid name + valid namespace/name ref listing

gabemontero:
a second case could also be added @jitendar-singh - creating a shared secret named `openshift-etc-pki-entitlement` that points to a secret other that the `etc-pki-entitlement` secret in the `openshift-config-managed` namespace

that should produce an error

jitendar-singh:
@gabemontero I have added the scenarios as required. Please review

gabemontero (approved):
/approve
/lgtm

gabemontero:
@jitendar-singh - just in the last couple of hours, there has been hypershift testing which has uncovered items which might forces us to go back to tech preview

I *don't think* that alters what you dropped here, but let me know what you think.  

And certainly based on when you sign off on https://issues.redhat.com/browse/BUILD-479 which 4.13 level you use may require enabling tech preview.

PR #130, opened 2023-03-13
OCPBUGS-10022: add 'system:serviceaccounts' to SAR to allow group based RBAC

gabemontero:
see https://redhat-internal.slack.com/archives/C014MHHKUSF/p1674829936947939


gabemontero:
fyi @sreber84

gabemontero:
/hold

@sreber84 is still sorting out some rbac grouping best practices with the auth team

gabemontero:
/hold cancel

no other work around coming

we'll include `system:serviceaccounts` in all SARs, but for other groups, an RFE, OCP EP, and API change for cluster admins to specify which groups to try the SAR with would be required.

@sreber84 FYI

PR #165, opened 2024-01-30
OCPBUGS-28663: Replace 'coreydaley' with 'sayan-biswas' in OWNERS file

coreydaley:


gabemontero:
> The following users are mentioned in OWNERS file(s) but are untrusted for the following reasons. One way to make the user trusted is to add them as [members](https://github.com/orgs/openshift/people) of the openshift org. You can then trigger verification by writing `/verify-owners` in a comment.
> 
>     * sayan-biswas
>       
>       * User is not a member of the org. User is not a collaborator. Satisfy at least one of these conditions to make the user trusted.
>       * OWNERS

I'm assuming you caught this ^^ @coreydaley .... will need to sort that out before this merges

/lgtm

coreydaley:
/verify-owners
/assign @adambkaplan 
for lgtm

PR #173, opened 2024-05-06
BUILD-957: Update Images with Red Hat Catalog Images

avinal:
- OpenShift CI registry is not available in Konflux and using secret to
  access is not a good idea
- updated Dockerfiles to use Red Hat catalog images: https://catalog.redhat.com/software/containers/explore
- Refer [BUILD-957](https://issues.redhat.com//browse/BUILD-957)

Signed-off-by: Avinal Kumar <avinal@redhat.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.21-openshift-4.16 AS builder
+FROM registry.access.redhat.com/ubi9/go-toolset:1.20.12 AS builder
 WORKDIR /go/src/github.com/openshift/csi-driver-shared-resource
 # to make SAST/SNYK happy
 RUN rm -rf examples
@@ -8,7 +8,7 @@ RUN rm -rf /go/src/github.com/openshift/csi-driver-shared-resource/examples
 RUN rm -f /go/src/github.com/openshift/csi-driver-shared-resource/vendor/k8s.io/apimachinery/pkg/util/managedfields/pod.yaml
 RUN make build
 
-FROM registry.ci.openshift.org/ocp/4.16:base-rhel9
+FROM registry.access.redhat.com/ubi9/ubi:9.4
 COPY --from=builder /go/src/github.com/openshift/csi-driver-shared-resource/_output/csi-driver-shared-resource /usr/bin/
 ENTRYPOINT []
 CMD ["/usr/bin/csi-driver-shared-resource"]
--- a/Dockerfile.webhook
+++ b/Dockerfile.webhook
@@ -1,4 +1,4 @@
-FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.21-openshift-4.16 AS builder
+FROM registry.access.redhat.com/ubi9/go-toolset:1.20.12 AS builder
 WORKDIR /go/src/github.com/openshift/csi-driver-shared-resource
 # to make SAST/SNYK happy
 RUN rm -rf examples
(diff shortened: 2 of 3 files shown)


Dockerfile.webhook
```diff
@@ -8,6 +8,6 @@ RUN rm -rf /go/src/github.com/openshift/csi-driver-shared-resource/examples
 RUN rm -f /go/src/github.com/openshift/csi-driver-shared-resource/vendor/k8s.io/apimachinery/pkg/util/managedfields/pod.yaml
 RUN make build-webhook
 
-FROM registry.ci.openshift.org/ocp/4.16:base-rhel9
+FROM registry.access.redhat.com/ubi9:9.3
```
sayan-biswas, line 11: Can't we use the ubi9-minimal?

avinal: Yes, this should be minimal, let me replace.

Makefile
```diff
@@ -28,7 +28,7 @@ ifeq ($(TARGET_GOARCH), amd64)
 	RACE = -race
 endif
 
-GOFLAGS ?= -a -mod=vendor $(RACE)
+GOFLAGS ?= -a -mod=vendor -buildvcs=false $(RACE)
```
avinal, line 31: I have added this option otherwise, the go compiler complains about missing VCS info: https://stackoverflow.com/questions/76997566/error-obtaining-vcs-status-exit-status-128-use-buildvcs-false-to-disable-vcs-s

Dockerfile
```diff
@@ -8,7 +8,7 @@ RUN rm -rf /go/src/github.com/openshift/csi-driver-shared-resource/examples
 RUN rm -f /go/src/github.com/openshift/csi-driver-shared-resource/vendor/k8s.io/apimachinery/pkg/util/managedfields/pod.yaml
 RUN make build
 
-FROM registry.ci.openshift.org/ocp/4.16:base-rhel9
+FROM registry.access.redhat.com/ubi9/ubi-minimal:9.4
```
adambkaplan, line 11: I think this needs to be updated to `ubi:9.4` - tests failing due to missing the `mount` command:

```
        E0506 11:02:26.024479       1 server.go:118] GRPC error: rpc error: code = Internal desc = failed to mount device: e2e-test-csi-driver-shared-resource-8ksnv-e2e-test-csi-driver-shared-resource-8ksnv-csi-312ee9d4ec3256014f4706131c4ce8a572cfa299188ddad473c67b7c61eeaa16 at /var/lib/kubelet/pods/5fe676ed-9792-4f2d-a969-617a9a28b8d3/volumes/kubernetes.io~csi/my-csi-volume/mount: mount failed: exec: "mount": executable file not found in $PATH
        Mounting command: mount
```

Makefile
adambkaplan: Is this go compiler error happening on Konflux, or OpenShift CI?

Note that the linked question on Stackoverflow does not have an accepted answer.

adambkaplan (changes requested):

Makefile
avinal: It is happening in Konflux.

adambkaplan:
/retitle BUILD-957: Update Images with Red Hat Catalog Images

Makefile
sayan-biswas: This is not happening in OpenShift CI, only in Konflux. OpenShift CI is using a different builder image, maybe it because of image we are using Konflux.

adambkaplan:
I ran into the same issue with source-to-image when I tried to include the `.git` directory. If you add `.git` to the `.dockerignore` file, the issue goes away. This is more likely a problem with Red Hat's `go-toolset:1.20` image.

adambkaplan (approved):
/approve

I was able to reproduce the `-buildvcs` issue in a separate repo. Setting that flag explicitly to false is fine for now.

adambkaplan:
/label docs-approved

/label px-approved

This has no impact on docs or product experience.

PR #174, opened 2024-05-06
BUILD-953: Simplify README

adambkaplan:
Simplifying the README so it is easier for upstream communities to understand and consume.

- Simplify "Getting Started" to recommend deploying OpenShift with tech preview features enabled.
- Add "How to Use" section with simple example.
- Simplify features list language.
- List CSI interfaces implemented.
- Move local development section to its own file, and link to it from README

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,77 +1,138 @@
 # OpenShift Shared Resource CSI Driver
 
-The OpenShift Shared Resource CSI Driver allows for the controlled (via Kubernetes RBAC) sharing of Kubernetes Secrets and ConfigMaps across 
-Namespaces in Openshift.
+The OpenShift Shared Resource CSI Driver allows `Secrets` and `ConfigMaps` to be
+shared across Kubernetes namespaces in a controlled manner. This CSI driver
+ensures that the entity (ServiceAccount) accessing the shared Secret or ConfigMap
+has permission to do so before mounting the data as a volume into the requesting
+Pod.
 
-The API used to achieve this support are:
+This CSI driver only supports the `Ephemeral` volume lifecycle mode.
+It also requires the following during operation:
 
-- the `volume` and `volumeMount` fields of a Kubernetes Pod
-- a new `SharedConfigMap` Kubernetes Custom Resource Definition which specifies which ConfigMap is to be shared, and which
-serves as the resource in Kubernetes Subject Access Review checks 
-- a new `SharedSecret` Kubernetes Custom Resource Definition which specifies which Secret is to be shared, and which
-  serves as the resource in Kubernetes Subject Access Review checks
-  
-## Features
-
-- Supports only a subset of the Kubernetes CSIVolumeSource API.  See [CSI Volume Specifics](docs/csi.md) for details.
-- Initial pod requests for `SharedConfigMap` or `SharedSecret` CSI volumes are denied without both a valid `SharedConfigMap` or `SharedSecret` reference and
-  permissions to access that `SharedConfigMap` or `SharedSecret`.
-- Changes to the `SharedConfigMap` or `SharedSecret` backing resource (namespace, name) get reflected in data stored in the user pod's CSI volume.
-- Subsequent removal of permissions for a `SharedConfigMap` or `SharedSecret` results in removal of the associated data stored in the user pod's CSI volume.
-- Re-granting of permission for a `SharedConfigMap` or `SharedSecret` (after having the permissions initially, then removed) results in the associated
-  data getting stored in the user pod's CSI volume.
-- Removal of the `SharedConfigMap` or `SharedSecret` used to provision a `SharedConfigMap` or `SharedSecret` csi volume for a pod results in the associated data getting removed.  
-- Re-creation of a removed `SharedConfigMap` or `SharedSecret` for a previously provisioned `SharedConfigMap` or `SharedSecret` CSI volume results in the associated data
-  reappearing in the user pod's CSI volume.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,77 +1,138 @@
...
-- Re-creation of a removed `SharedConfigMap` or `SharedSecret` for a previously provisioned `SharedConfigMap` or `SharedSecret` CSI volume results in the associated data
-  reappearing in the user pod's CSI volume.
-- Supports recycling of the csi driver so that previously provisioned CSI volumes are still managed; in other words,
-  the driver's internal state is persisted.
-- Multiple `SharedResources` within a pod are allowed.
-- When multiple `SharedResources` are mounted in a pod, one `SharedConfigMap` or `SharedSecret` can be mounted as a subdirectory of another `SharedConfigMap` or `SharedSecret`.
-
-
-NOTE: see [CSI Volume Specifics](docs/csi.md) for restrictions around these features for read-only Volumes.
+- `podInfoOnMount: true`
```
gabemontero, line 12: ah from the operator, since it owns the driver definition ... cool

README.md
```diff
@@ -1,77 +1,138 @@
...
+   metadata:
+     name: example-shared-config
+     namespace: app-namespace
+   spec:
+     ...
+     serviceAccountName: default
+     volumes:
+       - name: shared-config
+         csi:
+           readOnly: true
```
gabemontero, line 97: maybe we should put a yaml comment here that `readOnly: true` is required

README.md
```diff
@@ -1,77 +1,138 @@
...
+  [FAQ](docs/faq.md) for more details.
+- Automatic sync of the shared resource data (Secret/ConfigMap) into the mounting
+  `Pod`.
+- Automatic removal/restoration of shared resource data if the Pod's RBAC
+  permissions change at runtime.
+- Automatic removal/restoration of shared resource data if the backing
+  Secret/ConfigMap is deleted/re-created.
+- Survival of shared resource data with CSI driver restarts/upgrades.
+- Multiple `SharedSecret`/`SharedConfig` volumes within a `Pod`. Also supports
+  nested volume mounts within a container.
```
gabemontero, line 122: IMO if we don't say something about readOnly in that yaml, we should mention it here.

Also, although it was originally openshift motivated, I think the list of allowed names feature for shared secret/configmap should be listed too.

README.md
```diff
@@ -1,77 +1,138 @@
...
+
+The following CSI interfaces are implemented:
+
+- **Identity Service**: GetPluginInfo, GetPluginCapabilities, Probe
+- **Node Service**: NodeGetInfo, NodeGetCapabilities, NodePublishVolume, NodeUnpublishVolume
```
gabemontero, line 127: highlight this is a partial implementation of the NodeService interface

docs/local-development.md
```diff
@@ -0,0 +1,15 @@
...
+where you set the `DRIVER_IMAGE` environment variable to the image reference created by your `make build-image` against
+your local clone of this repository.
+
+See that operator's [quick start guide](https://github.com/openshift/csi-driver-shared-resource-operator#quick-start) for 
+complete details.
+
+NOTE: changes to API objects that act in concert with the driver (RBAC, CSI Driver definition, service, serviceaccounts, etc)
+are defined at [https://github.com/openshift/csi-driver-shared-resource-operator/tree/master/assets](https://github.com/openshift/csi-driver-shared-resource-operator/tree/master/assets).
+If your changes need adjustments to those objects, you'll need to use `make deploy` to rollout a new version of the operator,
+per the same quick start guide.
```
gabemontero, line 15: I wonder if the "RHTAP short cut" we devised to bypass tech preview and the operator is worth curating as part of upstreaming this @adambkaplan 

i.e. just apply the yaml noted in our old kustomization file (which has been deleted in the latest version):  https://github.com/gabemontero/infra-deployments/blob/d019b9be148dfa6fbf4ce4d0980389484dfd0ba5/components/shared-resources/kustomization.yaml

@sayan-biswas @avinal FYI

gabemontero (commented):
looks good @adambkaplan 

a few things did occur to me ... let's see what you think about them.

sayan-biswas (approved):

docs/local-development.md
adambkaplan: For upstreaming a Helm chart makes sense. But we can wait and see what the SIG requests/requires.

README.md
gabemontero: so did you miss this @adambkaplan or did you not agree with adding the bits about readOnly and reserved names?

README.md
gabemontero: similarly, did you miss this, or not agree with citing the methods that return unimplemented @adambkaplan ?

docs/local-development.md
gabemontero: sounds good .... just something to keep in mind / in our back pocket

README.md
adambkaplan: I think just leaving as-is is fine for now. I was trying to emulate what other kubernetes-sigs CSI drivers published in their READMEs.

README.md
adambkaplan: I'll add a follow-up on the reserved names - that is definitely a feature we need to call out.

For `readOnly` - ditto will add a comment in my YAML example.

README.md
gabemontero: sounds good thanks for confirming

README.md
gabemontero: sounds good thanks

PR #176, opened 2024-05-07
BUILD-953: Compare to Other Secret Stores

adambkaplan:
Expand the FAQ so this CSI driver can be compared to other secret store providers in the cloud-native ecocsystem (External Secrets Operator and Secret Store CSI driver). This will help other communities and end users evaluate this CSI driver in the context of similar tools.

Also added a few items missing from the README simplification:

- Indicate that `readOnly: true` is required for the volume definition.
- Add "reserved names" feature to the README.
- Add TODO to make the reserved names feature upstream-friendly. Currently a reserved name must have the "openshift" prefix to work. This might be a bug in the implementation.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -95,3 +95,48 @@ in real time.
 
 Conversely, if the kubelet is still in a retry cycle trying to launch a Pod with a `SharedConfigMap` or `SharedSecret` reference, if now resolved permission issues were the only thing preventing
 a mount, the mount should then succeed.  Of course, as kubelet retry vs. controller re-list is the polling mechanism, and it is more frequent, the change in results would be more immediate in this case.
+
+# Other Secret Providers/Operators
+
+The Shared Resource CSI driver has similar features and technical capabilities as
+other "secrets operator" projects. This CSI driver is complementary to many of these
+projects, especially those designed to fetch secrets from external providers such as
+HashiCorp Vault, AWS, Azure, and other cloud provider secret storage.
+
+## Can't I use the External Secrets Operator to share resources in my cluster?
+
+For Secrets, yes. The [External Secrets Operator](https://external-secrets.io) can
+share Secrets through its [Kubernetes provider](https://external-secrets.io/latest/provider/kubernetes/).
+However, there are several technical differences between this and External Secrets
+operator - see the "Feature Comparison" section below.
+
+## Can I share resources with the Secret Store CSI Driver?
+
+Not at present. Like this project, the [Secret Store CSI Driver](https://secrets-store-csi-driver.sigs.k8s.io)
+allows secrets to be synced and mounted with `CSI` volumes. A separate [provider](https://secrets-store-csi-driver.sigs.k8s.io/providers)
+is responsible for fetching and syncing the referenced data. As of this writing, no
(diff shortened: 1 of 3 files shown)


gabemontero:
> Expand the FAQ so this CSI driver can be compared to other secret store providers in the cloud-native ecocsystem (External Secrets Operator and Secret Store CSI driver). This will help other communities and end users evaluate this CSI driver in the context of similar tools.
> 
> Also added a few items missing from the README simplification:
> 
>     * Indicate that `readOnly: true` is required for the volume definition.
> 
>     * Add "reserved names" feature to the README.
> 
>     * Add TODO to make the reserved names feature upstream-friendly. Currently a reserved name must have the "openshift" prefix to work. This might be a bug in the implementation.

Yep the `openshift` prefix was intentional in the "openshift only" context.

For upstreaming, either allowing the configuration of required prefix, or allowing the definition of a regexp, probably makes sense.

docs/faq.md
```diff
@@ -95,3 +95,38 @@ in real time.
...
+## Can this be accomplished with the Secret Store CSI Driver?
+
+Not at present. The [Secret Store CSI Driver](https://secrets-store-csi-driver.sigs.k8s.io)
+also allows secrets to be synced and mounted with `CSI` volumes. A separate [provider](https://secrets-store-csi-driver.sigs.k8s.io/providers)
+is responsible for fetching and syncing the referenced data. As of this writing, no
+provider syncs data that explicitly exists within a Kubernetes cluster.
+
+Early in this project's history, we considered implementing a provider for the
+Secret Store CSI Driver. This was declined in favor the current RBAC model and
+implementing the CSI interfaces required for `Ephemeral` volume lifecycle mode.
```
gabemontero, line 117: We often talked about the Secret Store and Shared resources being complementary:

- use Secret Source to bring in the credentials from a source outside the Kubernetes cluster, but into one namespace

-   use Shared Secrets to allow other users to leverage the credentials brought in by Secret Store, without having to copy the secret, or define a Secret Store in all the namespaces that need those credentials.

IMO this complimentary relationship exists regardless of whether we ever implemented a provider for Secret Store.

pkg/config/namewatcher.go
```diff
@@ -67,6 +67,7 @@ func startsWithOpenShift(shareName string) bool {
 }
 
 func innerValidate(shareName, refNamespace, refName string, ok bool, v types.NamespacedName) bool {
+	// TODO: Make this configurable, do not skip reservation logic if resource share name does not start with "openshift*".
```
gabemontero, line 70: +1

gabemontero (commented):
just one minor potential add

docs/faq.md
adambkaplan: Good points - I tried to address as follows:

- Adding an intro paragraph in the section for the related Secret Store and External Secrets Operator projects
- Adding third FAQ - "Can I deploy this alongside the Secret Store CSI Driver?"

gabemontero (approved):
/lgtm

adambkaplan:
/label docs-approved

/label px-approved

/label qe-approved

No change to code - this is upstream documenation only.

PR #180, opened 2024-05-13
BUILD-961: use separate dockerfile for using catalog images

avinal:
- Konflux does not support using OpenShift CI images hence use a
  modified dockerfile for Konflux

Signed-off-by: Avinal Kumar <avinal@redhat.com>

### DIFF
--- a/.konflux/cdsr-webhook/Dockerfile
+++ b/.konflux/cdsr-webhook/Dockerfile
@@ -1,0 +1,10 @@
+FROM registry.access.redhat.com/ubi9/go-toolset:1.20.12 AS builder
+WORKDIR /go/src/github.com/openshift/csi-driver-shared-resource
+COPY . .
+RUN rm -rf /go/src/github.com/openshift/csi-driver-shared-resource/examples
+RUN rm -f /go/src/github.com/openshift/csi-driver-shared-resource/vendor/k8s.io/apimachinery/pkg/util/managedfields/pod.yaml
+RUN make build-webhook
+
+FROM registry.access.redhat.com/ubi9/ubi:9.4
+COPY --from=builder /go/src/github.com/openshift/csi-driver-shared-resource/_output/csi-driver-shared-resource-webhook /usr/bin/
+ENTRYPOINT ["/usr/bin/csi-driver-shared-resource-webhook"]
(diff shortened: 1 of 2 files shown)


.konflux/cdsr-webhook/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM registry.access.redhat.com/ubi9/go-toolset:1.20.12 AS builder
+WORKDIR /go/src/github.com/openshift/csi-driver-shared-resource
+# to make SAST/SNYK happy
+RUN rm -rf examples
+RUN rm -f vendor/k8s.io/apimachinery/pkg/util/managedfields/pod.yaml
```
adambkaplan, line 5: Why is this here? There should be no content at this stage of the build.

.konflux/cdsr-webhook/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+COPY . .
+RUN rm -rf /go/src/github.com/openshift/csi-driver-shared-resource/examples
+RUN rm -f /go/src/github.com/openshift/csi-driver-shared-resource/vendor/k8s.io/apimachinery/pkg/util/managedfields/pod.yaml
```
adambkaplan, line 8: I would rather us use a `.dockerignore` file at the project root - but we can address in a follow-up pull request.

https://docs.docker.com/build/building/context/#dockerignore-files

.konflux/cdsr/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM registry.access.redhat.com/ubi9/go-toolset:1.20.12 AS builder
+WORKDIR /go/src/github.com/openshift/csi-driver-shared-resource
+# to make SAST/SNYK happy
+RUN rm -rf examples
+RUN rm -f vendor/k8s.io/apimachinery/pkg/util/managedfields/pod.yaml
```
adambkaplan, line 5: Ditto - these instructions won't do anything at this stage of the build.

adambkaplan (approved):
/lgtm

Giving this the :+1: since it is copying over the existing Dockerfile content. However I think there is area for improvement as follow-up.

adambkaplan:
/label docs-approved

/label px-approved

@avinal this should be linked to the appropriate JIRA issue.

adambkaplan (approved):
/lgtm

adambkaplan:
/label qe-approved

No impact on QE

adambkaplan:
/label acknowledge-critical-fixes-only

This is a zero-risk change to begin onboarding this component to Konflux. Eventually we will remove the shared resource CSI driver from the OpenShift payload.

PR #226, opened 2024-11-03
Update golang version from 1.21 to 1.22

sayan-biswas:
Update golang version from 1.21 to 1.22


sayan-biswas:
Overriding konflux test temporarily and merging since konflux test requires new pipelines.

PR #351, opened 2025-04-11
[Builds-1.2] Added Pipeline as reference and updated images

psrvere:
# Changes

- Updated ubi and go-toolset images
- Update pipeline as reference


sayan-biswas (approved):

avinal:
It is preferable to do only one task in a PR. It is fine for this one.

avinal (approved):

PR #374, opened 2025-06-30
fix: change image in dockerfile to remove setuptools package

anchi205:
In .konflux/shared-resource/Dockerfile
The image we were using had the vulnerable package  (which we are not using)  

The change includes the use of UBI minimal image which doesn't have the setuptools package.


sayan-biswas (changes requested):
Can you rather a use latest tag? like "9.6"

Also the webhook container also needs this change. since the that is also a CVE.

avinal:
@anchi205 please mention the CVE in the pull request description, and a JIRA if applicable, in the title. See this for example openshift/jenkins#1836

sayan-biswas (approved):

PR #375, opened 2025-06-30
chore: use image tags instead of SHA

avinal:
- updated images for CSI Driver and CSI Driver webhook
- changed SHA to image tags


avinal:
Strange, closing in favor of #374

PR #420, opened 2025-07-11
Add mount binary to base image

sayan-biswas:
## Changes
- Mount binary added to base image by installing package util-linux
- Builder image updated to go 1.24


sayan-biswas:
The PR will also need to the package `util-linux` in RPM lock file.

avinal (approved):
Looks good to me

PR #430, opened 2025-08-08
[BUILD-984] Add test to verify RBAC checks

anchi205:
This PR adds a tests which ensure mount fails when csi-driver or pod lacks required permissions to access SharedSecrets/ConfigMaps.

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive RBAC configuration guide covering driver-level and pod-level permissions required for mounting shared secrets and config maps, including troubleshooting scenarios.

* **Bug Fixes**
  * Fixed potential nil reference when a shared secret is no longer available.

* **Tests**
  * Added end-to-end RBAC tests validating permission scenarios for shared secrets and config maps.
  * Enhanced test framework with improved setup and shared resource helpers.

* **Chores**
  * Updated build environment to Go 1.25 and OpenShift 4.21.

### DIFF
--- a/pkg/csidriver/driver.go
+++ b/pkg/csidriver/driver.go
@@ -599,6 +599,10 @@ func mapBackingResourceToPod(dv *driverVolume) error {
 		// we can return the error back to volume provisioning, where the kubelet will retry at
 		// a controlled frequency
 		sharedSecret := client.GetSharedSecret(dv.GetSharedDataId())
+		if sharedSecret == nil {
+			klog.V(4).Infof("mapBackingResourceToPod for pod volume %s:%s:%s share %s no longer exists", dv.GetPodNamespace(), dv.GetPodName(), dv.GetVolID(), dv.GetSharedDataId())
+			return nil
+		}
 		sNamespace := sharedSecret.Spec.SecretRef.Namespace
 		sName := sharedSecret.Spec.SecretRef.Name
 		comboKey := objcache.BuildKey(sNamespace, sName)
--- a/.ci-operator.yaml
+++ b/.ci-operator.yaml
@@ -1,4 +1,4 @@
 build_root_image:
   name: release
   namespace: openshift
-  tag: rhel-9-release-golang-1.24-openshift-4.20
+  tag: rhel-9-release-golang-1.25-openshift-4.21
--- a/README.md
+++ b/README.md
@@ -154,6 +154,7 @@ Sharing resources is done as follows:
 
 See also:
 
+- [RBAC configuration](docs/rbac.md)
 - [Simple example](docs/simple-example.md)
 - [Tekton example](docs/tekton-example.md)
 - [OpenShift BuildConfig example](docs/build-with-rhel-entitlements.md)
--- a/docs/rbac.md
+++ b/docs/rbac.md
@@ -1,0 +1,213 @@
+# RBAC Configuration for Shared Resources
+
+The Shared Resource CSI Driver enforces two layers of RBAC before mounting a
+`SharedSecret` or `SharedConfigMap` into a Pod:
+
+1. **Driver-level RBAC** — The CSI driver's ServiceAccount must have
+   `get`, `list`, and `watch` permissions on the source Secret or ConfigMap in
+   its namespace.
+2. **Pod-level RBAC** — The Pod's ServiceAccount must have `use` permission on
+   the `SharedSecret` or `SharedConfigMap` resource.
+
+Both layers must be satisfied. If either is missing, the volume mount fails and
+the Pod stays in a pending state with a `FailedMount` event.
+
+## Prerequisites
+
+- OpenShift cluster with Shared Resource CSI Driver deployed 
+- The CSI driver DaemonSet runs under the `csi-driver-shared-resource`
+  ServiceAccount in the `openshift-builds` namespace 
+
+## Granting Driver Permission to Access the Source Resource
+
+The CSI driver no longer has cluster-wide read access to all Secrets and
+ConfigMaps. You must explicitly grant it permission to read each source
+resource by creating a `Role` and `RoleBinding` in the source namespace.
+
+### For a Secret
+
+```yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: csi-read-my-secret
+  namespace: source-namespace
+rules:
+  - apiGroups: [""]
+    resources: ["secrets"]
+    resourceNames: ["my-secret"]
+    verbs: ["get", "list", "watch"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: csi-read-my-secret
+  namespace: source-namespace
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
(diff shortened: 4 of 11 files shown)


pkg/csidriver/driver.go
```diff
@@ -441,13 +443,14 @@ func (r *innerShareUpdateRanger) Range(key, value interface{}) bool {
 			sharedSecret := client.GetSharedSecret(r.shareId)
 			if sharedSecret == nil {
 				klog.Warningf("innerShareUpdateRanger unexpected not found on sharedSecret lister refresh: %s", r.shareId)
-				return false
+				return true // if the driver fails to get the source `Secret` for the 3rd volume, it logs a warning and returns `true`. This allows the loop to continue on to the 4th, 5th, and all subsequent volumes, updating them correctly.
```
sayan-biswas, line 446: Why do we want to change this behaviour?

pkg/csidriver/driver.go
```diff
@@ -547,19 +551,27 @@ func mapBackingResourceToPod(dv *driverVolume) error {
...
-			if upsertError != nil {
-				ProcessFileSystemError(cm, upsertError)
-				return upsertError
+		cm, err := client.GetConfigMap(cmNamespace, cmName)
+		if err != nil { // If the error is `Permission Denied, it returns a `PermissionDenied` error, which correctly fails the initial volume mount
+			if kerrors.IsForbidden(err) {
+				return status.Errorf(codes.PermissionDenied, "CSI driver does not have permission to get configmap %s/%s", cmNamespace, cmName)
 			}
+			klog.Warningf("could not retrieve configmap %s/%s: %v", cmNamespace, cmName, err)
+			return nil
```
sayan-biswas, line 561: Are these change in execution required for the e2e test?
I feel these changes should be captured in a different PR

sayan-biswas (commented):
Context segregation required. These changes the workflow of the shared resource mounts. What is the motivation behind these changes?

pkg/csidriver/driver.go
gabemontero: First, the comment references `Secret` when it should reference `SharedSecret`.

But if I am understanding the flow correctly after revisiting the code, I think the premise in the comment is false as well.  And hence agree with the motivation for @sayan-biswas 's question here, that we should not change the behavior here.

But if you have say logs from a test run @anchi205 that say differently, please provide.

To explain, some tl;dr

So, given the explanation of the golang `Range` function:  
```
Explanation:
m.Range(func(key, value interface{}) bool { ... }):
The Range method takes a function as an argument. This function is called for each key-value pair in the map.
key, value interface{}:
The key and value parameters within the anonymous function are of type interface{}, meaning you need to perform type assertions if you want to work with their concrete types (e.g., key.(string)).
return true:
If the function returns true, the iteration continues to the next key-value pair.
return false:
If the function returns false, the iteration stops immediately. This allows you to control the iteration flow based on your logic, for example, stopping once a specific condition is met.
```

I could see how it would seem that returning false means it would not process other volumes in the `volumes` Map in driver.go

But I think the confusion stems from the fact that we are only expecting on entry to be processed.  Hence there is no need to return `true` and process the next one.

Consider `shareUpdateRanger` over at https://github.com/openshift/csi-driver-shared-resource/blob/3f665fd9d18924c11c9adffdf8d07e10baf380cc/pkg/csidriver/driver.go#L496-L514

We set the `shareId` in the `innerShareUpdateRanger` object.  And then the vast majority of the code is only run when the `driveVolume` corresponds to that one and only one `shareId`.  There is only one `driveVolume` that will match that `shareId`.

If an error ocurs, returning `false` just speeds up the processing.  Since only one `driveVolume` is the correct one, none of the other `driveVolumes` from the `volumes` map will match.

gabemontero: so returning `true` just means iterating on other `driveVolumes` that will never match the `shareId`

pkg/cache/configmaps.go
```diff
@@ -69,12 +69,12 @@ func RegisterConfigMapUpsertCallback(volID, cmID string, f func(key, value inter
 	}
 	configmapUpsertCallbacks.Store(volID, f)
 	ns, name, _ := SplitKey(cmID)
-	cm := client.GetConfigMap(ns, name)
-	if cm != nil {
-		f(BuildKey(cm.Namespace, cm.Name), cm)
-	} else {
-		klog.Warningf("not found on get configmap for %s vol %s", cmID, volID)
+	cm, err := client.GetConfigMap(ns, name)
```
gabemontero, line 72: I do think this makes sense now that https://issues.redhat.com/browse/BUILD-965 means the driver might not be able to read the secret

In fact, I would change the error signature of this method to return and error, and propagate that error up the call stack

Seems like that was missed when https://issues.redhat.com/browse/BUILD-965 was implemented

pkg/csidriver/driver.go
gabemontero: Maybe the need here was the kubelet needs those GRPC error codes vs. what is in the k8s forbidden exception?

If so, then I could see how you would want to return permission denied here 

but I would not reutrn nil if it is a non permission error .... either return the original error like before, or find the generic GRPC error code and do a status.Errorf... with that

pkg/csidriver/driver.go
```diff
@@ -589,21 +601,33 @@ func mapBackingResourceToPod(dv *driverVolume) error {
...
-			if upsertError != nil {
-				ProcessFileSystemError(s, upsertError)
-				return upsertError
+		s, err := client.GetSecret(sNamespace, sName)
+		if err != nil { // If the error is `Permission Denied, it returns a `PermissionDenied` error, which correctly fails the initial volume mount
+			if kerrors.IsForbidden(err) {
+				return status.Errorf(codes.PermissionDenied, "CSI driver does not have permission to get secret %s/%s", sNamespace, sName)
 			}
+			klog.Warningf("could not retrieve secret %s/%s: %v", sNamespace, sName, err)
+			return nil
```
gabemontero, line 618: again don't eat this error and return nil ... return the error, or do another status.Errorf(codes.... with a generic error code

pkg/csidriver/driver.go
anchi205: This is my understanding on this. Please kindly correct me if I am wrong:

Multiple pods can mount the same share.  `createVolume` function is called for each new pod's mount request; it creates a new and unique `driverVolume` object for each pod but assigns them all the same `SharedDataId` if they are using the same share. This proves that many `driverVolume` entries can exist for a single `shareId`. 

`return true` is needed  because all the `SharedDataId`s are the same. The driver needs to find every single one of them.

Lets say a `Secret` is updated. It triggers the `UpsertSecret` function (loops through all `SharedSecret` objects to find the one that references this specific `Secret` ), which in turn calls the `shareUpdateRanger`. This function starts a `volumes.Range` loop which iterates over the `volumes` map, and each time, the key from that map ( volID - unique key for each pod's volume mount) is passed into the Range callback function. Range function casts this `key` to a string and names it `volID`, and then it filters all volumes that are using the `shareId` that was updated and then updates their data. 

Returning false would stop the update process for the remaining volumes which are using this share that references an updated Secret now.

sayan-biswas:
/hold

Release risk. (builds-1.6)

PR #431, opened 2025-08-13
BUILD-1545: Fix error-handling for shared resources RBAC

anchi205:
Files changed:

### 1. pkg/client/listers.go

- Functions Changed: `GetSecret`, `GetConfigMap`
- What Was Changed: The function signatures were modified to return an `error` object. The implementation was updated to capture and return any error received from the Kubernetes API client, instead of just logging it and returning `nil`.
    
- Why: This is the primary bug fix. The previous implementation ignored `Forbidden` errors, making it impossible for the driver to distinguish between a resource that was "not found" vs one it was denied access to. By propagating the error, this change enables the driver logic to correctly identify and act on the permission failures.

### 2. pkg/csidriver/driver.go

- Functions Changed: `mapBackingResourceToPod`, `innerShareUpdateRanger`
- What Was Changed:
    - In `mapBackingResourceToPod`, the logic was updated to handle the new `error` return from `GetSecret`/`GetConfigMap`. It now specifically checks if the error `IsForbidden` and, if so, returns the error to the caller. This fails the initial volume mount. A `nil` check was also added to prevent the driver from crashing if a non-forbidden error occurs.
    - In `innerShareUpdateRanger`, the logic was updated to handle the new `error` return. Additionally, in 2 places (when a resource fetch fails and when a pod's permissions are revoked), the function's return value was changed from `false` to `true`.
        
- Why:
    - The changes to `mapBackingResourceToPod` consume the bug fix. This is the logic that enforces the RBAC check during a pod's startup, preventing the security vulnerability.
    - Changing the return value to `true` ensures that if one volume update fails, the loop continues for all other volumes. This prevents a single failure from causing data inconsistency, which could lead to security risks.

### 3. pkg/cache/secrets.go and 4. pkg/cache/configmaps.go

- Functions Changed: `RegisterSecretUpsertCallback`, `RegisterConfigMapUpsertCallback`
- What Was Changed: The calls to `client.GetSecret` and `client.GetConfigMap` were updated to correctly handle the new return signature. The code now checks that the returned `error` is `nil` before proceeding to use the resource object.
    
- Why: As these functions were calling functions in `listers.go` (which were modified), these had to be updated.
    
### 5. pkg/cache/shares.go

- Functions Changed: `AddSharedConfigMap`, `AddSharedSecret`
- What Was Changed: The calls to client.GetConfigMap and client.GetSecret were updated to correctly handle the new return signature. The code now captures and logs any error that occurs when fetching the backing resource during a share creation event.

- Why: As these functions were calling functions in `listers.go` (which were modified), these had to be updated.

### 6. pkg/csidriver/driver_test.go

- Functions Changed: A new function, `TestMapBackingResourceHandlesForbiddenError`, was added.
- What Was Changed: A new table-driven unit test was added to provide specific coverage for the bug fix. This test uses a fake Kubernetes client and a reactor to simulate the API server returning a `Forbidden` error.
    
- Why: This test proves that the fix works for both Secrets and ConfigMaps. It asserts that when the driver encounters a `Forbidden` error, it correctly propagates that error instead of ignoring it.

### DIFF
--- a/pkg/cache/configmaps.go
+++ b/pkg/cache/configmaps.go
@@ -8,6 +8,9 @@ import (
 
 	"github.com/openshift/csi-driver-shared-resource/pkg/client"
 	"github.com/openshift/csi-driver-shared-resource/pkg/config"
+	"google.golang.org/grpc/status"
+	"google.golang.org/grpc/codes"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 )
 
 /*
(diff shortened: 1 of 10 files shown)


sayan-biswas:
/hold 

Release risk. (builds-1.6)

ayushsatyam146 (approved):
/lgtm

ayushsatyam146:
/override ci/prow/images ci/prow/okd-scos-images ci/prow/unit ci/prow/verify

sayan-biswas:
/override ci/prow/security ci/prow/unit ci/prow/images ci/prow/okd-scos-images ci/prow/verify

PR #505, opened 2026-04-02
BUILD-1921: correct callback unregistration to prevent memory leak in share-level maps

anchi205:
UnregisterSharedConfigMapUpdateCallback and UnregisterSharedSecretUpdateCallback
were deleting from the wrong sync.Map — configmapUpsertCallbacks and
secretUpsertCallbacks instead of shareConfigMapsUpdateCallbacks and
shareSecretsUpdateCallbacks.

Fix the two variable names so deleteVolume correctly unregisters the callbacks
it registered. Add callback count helpers to pkg/cache for testing. Add
TestCallbackLeakOnPodChurn, verifying all 8 callback maps are empty after
50 pod lifecycles. Fix missing deleteVolume call in
TestMapVolumeToPodWithKubeClient to prevent global sync.Map pollution
across tests.

## Release Notes

* **Bug Fixes**
  * Fixed callback cleanup issues during volume lifecycle operations to prevent potential resource leaks.
  * Corrected function naming inconsistency for shared resource callback management.

* **Tests**
  * Added validation test to ensure callbacks are properly cleaned up during repeated volume creation and deletion cycles.

### DIFF
--- a/pkg/cache/configmaps.go
+++ b/pkg/cache/configmaps.go
@@ -109,3 +109,15 @@ func RegisterConfigMapDeleteCallback(volID string, f func(key, value interface{}
 func UnregisterConfigMapDeleteCallback(volID string) {
 	configmapDeleteCallbacks.Delete(volID)
 }
+
+func ConfigMapUpsertCallbackCount() int {
+	count := 0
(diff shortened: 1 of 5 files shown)


sayan-biswas (approved):

sayan-biswas:
@anchi205 Can you check on the unit test?

PR #528, opened 2026-08-31
[BUILD-1841] feat(csi-driver): add TLS flag support and ML-KEM curves

anchi205:
Accept --tls-min-version and --tls-cipher-suites flags for webhook and metrics servers to enable cluster TLS profile adherence by the operator. Leave CurvePreferences unset to enable ML-KEM post-quantum cryptography support (Go 1.23+).

Changes:
- Add TLS flags to webhook and CSI driver
- Build tls.Config from flags
- Leave CurvePreferences unset for ML-KEM-768 support
- Add tests including TLS handshake validation

## Summary by CodeRabbit

* **New Features**
  * Added configurable TLS minimum versions and cipher suites for metrics and webhook HTTPS servers.
  * Added command-line flags for TLS settings.
  * TLS defaults to version 1.2 when no minimum version is specified.
  * Webhook HTTPS now supports configured client certificate validation.
* **Security**
  * Invalid TLS versions or cipher-suite lists now prevent startup with an error.
  * Unsupported cipher suites are skipped when valid alternatives are provided.
  * TLS 1.3 uses Go’s built-in cipher selection and curve preferences.
* **Reliability**
  * Added HTTP server timeouts and graceful metrics server shutdown.

### DIFF
--- a/cmd/csidriver/main.go
+++ b/cmd/csidriver/main.go
@@ -32,6 +32,8 @@ var (
 	driverName        string // name of the CSI driver, registered in the cluster
 	nodeID            string // current Kubernetes node identifier
 	maxVolumesPerNode int64  // maximum amount of volumes per node, i.e. per driver instance
+	tlsMinVersion     string // minimum TLS version for metrics server
+	tlsCipherSuites   string // comma-separated TLS cipher suites for metrics server
 
 )
 
@@ -46,7 +48,7 @@ var rootCmd = &cobra.Command{
 		cfgManager := config.NewManager(cfgFilePath)
 		cfg, err := cfgManager.LoadConfig()
 		if err != nil {
-			fmt.Printf("Failed to load configuration file '%s': %s", cfgFilePath, err.Error())
+			fmt.Printf("Failed to load configuration file '%s': %s\n", cfgFilePath, err.Error())
 			os.Exit(1)
 		}
 
@@ -55,13 +57,13 @@ var rootCmd = &cobra.Command{
 
 		}
 		if kubeClient, err := loadKubernetesClientset(); err != nil {
-			fmt.Printf("Failed to load Kubernetes API client: %s", err.Error())
+			fmt.Printf("Failed to load Kubernetes API client: %s\n", err.Error())
 			os.Exit(1)
 		} else {
 			client.SetClient(kubeClient)
 		}
 		if shareClient, err := loadSharedresourceClientset(); err != nil {
-			fmt.Printf("Failed to load SharedResource API client: %s", err.Error())
+			fmt.Printf("Failed to load SharedResource API client: %s\n", err.Error())
 			os.Exit(1)
 		} else {
 			client.SetShareClient(shareClient)
@@ -78,13 +80,13 @@ var rootCmd = &cobra.Command{
 			mount.New(""),
 		)
 		if err != nil {
-			fmt.Printf("Failed to initialize driver: %s", err.Error())
+			fmt.Printf("Failed to initialize driver: %s\n", err.Error())
 			os.Exit(1)
 		}
 
-		c, err := controller.NewController(cfg.GetShareRelistInterval(), cfg.RefreshResources)
+		c, err := controller.NewController(cfg.GetShareRelistInterval(), cfg.RefreshResources, tlsMinVersion, tlsCipherSuites)
 		if err != nil {
-			fmt.Printf("Failed to set up controller: %s", err.Error())
+			fmt.Printf("Failed to set up controller: %s\n", err.Error())
 			os.Exit(1)
 		}
 		prunerTicker := time.NewTicker(cfg.GetShareRelistInterval())
@@ -133,6 +135,8 @@ func init() {
 	rootCmd.Flags().StringVar(&driverName, "drivername", string(operatorv1.SharedResourcesCSIDriver), "name of the driver")
 	rootCmd.Flags().StringVar(&nodeID, "nodeid", "", "node id")
 	rootCmd.Flags().Int64Var(&maxVolumesPerNode, "maxvolumespernode", 0, "limit of volumes per node")
+	rootCmd.Flags().StringVar(&tlsMinVersion, "tls-min-version", "", "Minimum TLS version for metrics server (e.g., VersionTLS12)")
+	rootCmd.Flags().StringVar(&tlsCipherSuites, "tls-cipher-suites", "", "Comma-separated list of cipher suites for metrics server")
 }
 
 // loadKubernetesClientset instantiate a clientset using local config.
@@ -157,7 +161,7 @@ func loadSharedresourceClientset() (sharev1clientset.Interface, error) {
 func runOperator(c *controller.Controller, stopCh <-chan struct{}) {
 	err := c.Run(stopCh)
 	if err != nil {
-		fmt.Printf("Controller exited: %s", err.Error())
+		fmt.Printf("Controller exited: %s\n", err.Error())
 		os.Exit(1)
 	}
 }
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -8,12 +8,14 @@ import (
 	"net/http"
 	"os"
 	"strconv"
+	"time"
 
 	"github.com/spf13/cobra"
 
 	"k8s.io/klog/v2"
 
 	"github.com/openshift/csi-driver-shared-resource/pkg/config"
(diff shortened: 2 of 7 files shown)


cmd/webhook/main.go
```diff
@@ -63,33 +69,47 @@ func startServer() {
...
-		server.TLSConfig = &tls.Config{
-			RootCAs: certpool,
+		// Build TLS config from flags injected by operator
+		tlsCfg, err := tlsprofile.BuildTLSConfigFromFlags(tlsMinVersion, tlsCipherSuites)
+		if err != nil {
+			fmt.Printf("Invalid TLS configuration: %s\n", err.Error())
+			os.Exit(1)
 		}
+		tlsCfg.ClientCAs = certpool
+		tlsCfg.ClientAuth = tls.RequireAndVerifyClientCert
```
psrvere: This line makes the webhook require a client certificate on every connection. The kube-apiserver does not send one to admission webhooks by default, so every admission call would fail the TLS handshake.

I checked this with the webhook in a container. Without a client cert, curl fails and the server logs `tls: client didn't provide a certificate`. The base commit accepts the same request.

The operator passes `--cacert=/etc/pki/tls/certs/ca-bundle.crt`, the system CA bundle. That doesn't look like a client CA list. The old code put this pool in `RootCAs`, which a server never reads, so it had no effect. Could we keep that behaviour here and handle mTLS in a separate change?

cmd/webhook/main.go
```diff
@@ -63,33 +69,47 @@ func startServer() {
...
-		certpool.AppendCertsFromPEM(cafile)
+		if !certpool.AppendCertsFromPEM(cafile) {
+			fmt.Printf("Failed to parse CA certificate from %s\n", caCert)
+			os.Exit(1)
+		}
 
-		server.TLSConfig = &tls.Config{
-			RootCAs: certpool,
+		// Build TLS config from flags injected by operator
+		tlsCfg, err := tlsprofile.BuildTLSConfigFromFlags(tlsMinVersion, tlsCipherSuites)
```
psrvere: `:=` declares a new `err` inside the `if useTLS` block. The `err = server.ListenAndServeTLS(...)` below then writes to that inner variable. The `if err != nil` after the block reads the outer one, which is still nil.

I tested it with `--tlscert` pointing at a missing file. The base commit exits 1 and prints `Error serving connection`. This branch exits 0 and prints nothing.

Declaring `var tlsCfg *tls.Config` and assigning with `tlsCfg, err = ...` fixes it. I checked that with a test.

cmd/webhook/main.go
```diff
@@ -53,6 +57,8 @@ func init() {
 	CmdWebhook.Flags().StringVar(&listenAddress, "listen", "0.0.0.0", "Listen address")
 	CmdWebhook.Flags().IntVar(&listenPort, "port", 5000, "Secure port that the webhook listens on")
 	CmdWebhook.Flags().BoolVar(&testHooks, "testHooks", false, "Test webhook URI uniqueness and quit")
+	CmdWebhook.Flags().StringVar(&tlsMinVersion, "tls-min-version", "", "Minimum TLS version (e.g., VersionTLS12)")
+	CmdWebhook.Flags().StringVar(&tlsCipherSuites, "tls-cipher-suites", "", "Comma-separated list of cipher suites")
```
psrvere: Other Kubernetes components take IANA names on `--tls-cipher-suites`, such as `TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256`. This flag only takes OpenSSL names. With IANA names the webhook refuses to start with `no valid cipher suites found`, and the pod would crashloop.

Could the help text say which format it expects? Accepting both would also work. Which format will the operator send?

cmd/csidriver/main.go
```diff
@@ -133,6 +135,8 @@ func init() {
 	rootCmd.Flags().StringVar(&driverName, "drivername", string(operatorv1.SharedResourcesCSIDriver), "name of the driver")
 	rootCmd.Flags().StringVar(&nodeID, "nodeid", "", "node id")
 	rootCmd.Flags().Int64Var(&maxVolumesPerNode, "maxvolumespernode", 0, "limit of volumes per node")
+	rootCmd.Flags().StringVar(&tlsMinVersion, "tls-min-version", "", "Minimum TLS version for metrics server (e.g., VersionTLS12)")
+	rootCmd.Flags().StringVar(&tlsCipherSuites, "tls-cipher-suites", "", "Comma-separated list of cipher suites for metrics server")
```
psrvere: Same point as on the webhook flag. This only takes OpenSSL cipher names, while `--tls-cipher-suites` usually means IANA names in other components. Could the help text say which format it expects?

pkg/tls/profile.go
```diff
@@ -0,0 +1,91 @@
...
+		"AES128-SHA256":                 tls.TLS_RSA_WITH_AES_128_CBC_SHA256,
+	}
+
+	out := make([]uint16, 0, len(names))
+	for _, name := range names {
+		name = strings.TrimSpace(name)
+		if id, ok := m[name]; ok {
+			out = append(out, id)
+		} else {
+			klog.V(4).Infof("Cipher suite %q not supported by Go, skipping", name)
```
psrvere: A cipher that isn't mapped gets dropped with a `V(4)` log, which doesn't show at the default log level. I passed four ciphers, three of them SHA-1 or 3DES. Three were dropped and nothing was logged.

That's an admin's TLS profile being narrowed. A `klog.Warningf` would let them see it.

pkg/tls/profile.go
```diff
@@ -0,0 +1,91 @@
...
+			return nil, fmt.Errorf("invalid TLS version %q: %w", minVersionFlag, err)
+		}
+		cfg.MinVersion = minVer
+	} else {
+		// Safe default when not configured
+		cfg.MinVersion = tls.VersionTLS12
+	}
+
+	// Only set cipher suites for TLS 1.2 and below; Go hardcodes TLS 1.3 ciphers
+	if cfg.MinVersion < tls.VersionTLS13 && cipherSuitesFlag != "" {
```
psrvere: When the minimum version is TLS 1.3, the cipher list is ignored without a log line. Go picks TLS 1.3 ciphers itself, so skipping them is right. One info line would still help anyone reading the config back.

psrvere (commented):
I ran the webhook from this branch and from the base commit (`e3548bb4`) in podman containers, and probed them with curl and `openssl s_client`.

The TLS flags work. A TLS 1.3 minimum rejects TLS 1.2 clients, and a cipher outside the list is rejected.

Two problems in `cmd/webhook/main.go` show up in testing, see the inline comments. The webhook now rejects clients without a client certificate, which is how the kube-apiserver connects. A failed TLS listener now exits 0 with no error.

I reproduced both with a Go test in `cmd/webhook` that runs `startServer()` in a child process. Both cases pass on the base commit and fail on this branch. With the two fixes suggested inline, both pass. Happy to share the test if it's useful to add here.

pkg/controller/controller.go
```diff
@@ -116,7 +123,11 @@ func (c *Controller) Run(stopCh <-chan struct{}) error {
 
 	// start the Prometheus metrics serner
 	klog.Info("Starting the metrics server")
-	server, err := metrics.BuildServer(metrics.MetricsPort)
+	metricsTLSCfg, err := c.buildMetricsTLSConfig()
```
psrvere: Bad TLS flag values only come to light here, after the driver and informers are already running. Could `main` parse them once and pass a `*tls.Config` into `NewController`? The two string fields and `buildMetricsTLSConfig` could then go.

cmd/webhook/main.go
```diff
@@ -63,33 +69,47 @@ func startServer() {
...
 		os.Exit(0)
 	} else {
-		fmt.Printf("HTTP server running at: %s", net.JoinHostPort(listenAddress, strconv.Itoa(listenPort)))
+		fmt.Printf("HTTP server running at: %s\n", net.JoinHostPort(listenAddress, strconv.Itoa(listenPort)))
 	}
 
 	server := &http.Server{
-		Addr: net.JoinHostPort(listenAddress, strconv.Itoa(listenPort)),
+		Addr:         net.JoinHostPort(listenAddress, strconv.Itoa(listenPort)),
+		ReadTimeout:  30 * time.Second, // Kubernetes webhook timeout is 30s
```
psrvere: 30s is the maximum webhook timeout in Kubernetes. The default is 10s. The value is fine, but the comment reads as though 30s is always the timeout.

PR #529, opened 2026-09-01
[release-4.12] OCPBUGS-102157: CVE-2026-33814: replace golang.org/x/net with openshift-sustaining fork

mariocr73:
golang.org/x/net's real fix for CVE-2026-33814 (GO-2026-4918, malformed SETTINGS_MAX_FRAME_SIZE causing an infinite loop in golang.org/x/net/http2) ships in v0.53.0, which requires go>=1.25. This branch's go.mod declares go 1.18, so a direct module bump is not buildable here without also bumping the builder toolchain.

Uses `replace golang.org/x/net => github.com/openshift-sustaining/net v0.35.0-sec.3` (the sustaining team's own downstream fork) which backports only the CVE-2026-33814 fix without requiring a toolchain bump. Same mechanism already used by other components in this CVE (e.g. openshift/csi-operator, openshift/azure-disk-csi-driver-operator, openshift/route-controller-manager).

Verified: `go build ./...` and `go vet ./...` pass with this branch's real toolchain; `govulncheck` no longer reports GO-2026-4918 for this module.


avinal:
Hey @mariocr73, the security check has been broken since ever. Also, the project was moved as part of OpenShift Builds, which has the infrastructure on Konflux.

A doubt though, is 4.12 still maintained? Even if that is the case, AFIK, we do not have a release infra for these versions. Is that okay?

sayan-biswas (approved):

mariocr73:
> Hey @mariocr73, the security check has been broken since ever. Also, the project was moved as part of OpenShift Builds, which has the infrastructure on Konflux.
> 
> A doubt though, is 4.12 still maintained? Even if that is the case, AFIK, we do not have a release infra for these versions. Is that okay?

@avinal, Thanks for the heads-up! Yes, the security check issue is pre-existing — the failing tests here are all unrelated to this dependency bump (CI infra issues, not code changes).

Regarding 4.12 — it is still maintained by the Sustaining Engineering team, so this CVE fix is still needed. I understand there may not be active release infra on Konflux for this version, but the fix should still be merged for compliance.

Could we get the unrelated failures `/override`'d so this can move forward? Thanks!

